Issue #125, opened 2013-10-27
running on portable vim on windows


jmlucjav:
Hi,

I have tried vimgolfing on windows, but it fails to launch my vim. I use a portable version, and the exe that must be called is gVimPortable.exe. I have verified that if I take the commandline launched by vimgolf and replace 'vim.exe' by 'gVimPortable.exe' it runs ok.

I tried creating a hardlink vim.exe -> gVimPortable.exe but it also fails. Is there a way to configure the executable that must be invoked? I see this in the page;

```
Windows users: To compete you will need to make sure that you can launch "vim" and "diff" from your command prompt. The first one should be taken care of for you
```

so it sounds like maybe its possible? But I am not sure how.

Issue #100, opened 2012-11-25
Replace or explain "<fd-##>" and "<ff-##>" sequences


dbarnett:
Some solutions on vimgolf contain key sequences like `<ff-58>` and `<fd-35>` (see this Stack Overflow question: http://stackoverflow.com/questions/5225969/what-does-fd-35-mean-on-vimgolf). There's very little info anywhere on the internet about what that means.

If those are totally inert, vimgolf should filter them out of solutions so it's not confusing. If they're not totally inert and filtering them isn't an option, there should be some title text or link on vimgolf explaining what's going on.

PR #142, opened 2015-01-07
Play vimgolf on local files

urticadioica:
Introduces local play. You can take two local files and play Vimgolf on them. Good for:

* Testing challenges you're working on before submitting.
* Copy/pasting Vim strokes to forums and such. You can be sure the strokes you're posting actually work, and that there are no typos.
* Playing moar Vimgolf.

A few other changes had to be made to accommodate local play. They're all good changes anyway:

* The vimrc has to be included in the repository. (That's where it belongs anyway.) There's a security benefit too. It's possible for an attacker to intercept the json from vimgolf.com and put malicious stuff in the vimrc.
* Using temp files for work file and keylog.

I fixed another fairly severe security problem. Currently, the `@data['in']['type']` json field is taken straight from the Internet and stuck in three separate shell commands, unsanitized. An attacker can alter that field and shell-inject arbitrary code on the player's computer. The new version sanitizes the `@type` fields, and executes vim and vimdiff directly, without the shell. I simply replaced the diff call. (No more diff dependency!)

That and a few other things all ended up in 1 commit. Oh well.

### DIFF
--- a/lib/vimgolf/challenge.rb
+++ b/lib/vimgolf/challenge.rb
@@ -109,7 +109,7 @@ module VimGolf
     attr_reader :log_path
     attr_reader :vimrc_path
 
-    def correct
+    def correct?
       FileUtils.compare_file(@work_path, @output_path)
     end
 
--- a/lib/vimgolf/challenge.rb
+++ b/lib/vimgolf/challenge.rb
@@ -1,6 +1,6 @@
 module VimGolf
   class Challenge
-    attr_reader :id, :type
+    attr_reader :id, :type, :otype, :remote
 
     def self.path(path)
       @@path = path if path
@@ -11,7 +11,28 @@ module VimGolf
       @id = id
     end
 
+    def local(infile, outfile)
+      @remote = false
+      @input_path = File.expand_path(infile)
+      @output_path = File.expand_path(outfile)
+      @type = File.basename(@input_path) # extension? use the whole thing
+      @otype = File.basename(@output_path)
+
+      work_files
+    end
+
+    def work_files
+      @vimrc_path = File.expand_path('../vimgolf.vimrc', __FILE__)
+
+      # keep these temp files around so they don't close
+      @work = Tempfile.new(['vimgolf', ".#{@type}"])
+      @log = Tempfile.new('golflog')
+      @work_path = @work.path()
+      @log_path = @log.path()
+    end
+
     def download
+      @remote = true
       begin
         url = URI.parse("#{GOLFHOST}/challenges/#{@id}.json")
         req = Net::HTTP::Get.new(url.path)
@@ -34,9 +55,14 @@ module VimGolf
         @data['in']['data'].gsub!(/\r\n/, "\n")
         @data['out']['data'].gsub!(/\r\n/, "\n")
 
-        @type = @data['in']['type']
+        # be sure to sanitize the types
+        @type = @data['in']['type'].gsub(/[^\w-]/, '.')
+        @otype = @data['out']['type'].gsub(/[^\w-]/, '.')
+        @input_path = path + ".input.#{@type}"
+        @output_path = path + ".output.#{@otype}"
+
         save
-        start
+        work_files
       rescue Exception => e
         debug(e)
         raise "Uh oh, couldn't download or parse challenge, please verify your challenge id & client version."
@@ -44,13 +70,12 @@ module VimGolf
     end
 
     def start
-      File.open(work_path, "w")  {|f| f.puts @data['in']['data']}
+      FileUtils.cp(@input_path, @work_path)
     end
 
     def save
       File.open(input_path, "w")  {|f| f.puts @data['in']['data']}
       File.open(output_path, "w") {|f| f.puts @data['out']['data']}
-      File.open(vimrc_path, "w")  {|f| f.puts @data['vimrc']}
     end
 
     def upload
@@ -78,11 +103,15 @@ module VimGolf
       end
     end
 
-    def input_path;  path + ".#{@type}"; end
-    def work_path;   path + ".work.#{@type}"; end
-    def output_path; path + ".output"; end
-    def log_path;    path + ".log";    end
-    def vimrc_path;  path + ".golfrc"; end
+    attr_reader :input_path
+    attr_reader :work_path
+    attr_reader :output_path
+    attr_reader :log_path
+    attr_reader :vimrc_path
+
+    def correct
+      FileUtils.compare_file(@work_path, @output_path)
+    end
 
     def path
       @@path + "/#{@id}"
--- a/lib/vimgolf/cli.rb
+++ b/lib/vimgolf/cli.rb
@@ -120,7 +120,7 @@ module VimGolf
           VimGolf.ui.print_log log
 
           # Handle incorrect solutions
-          if !challenge.correct()
+          if !challenge.correct?()
             VimGolf.ui.error "\nUh oh, looks like your entry does not match the desired output."
             VimGolf.ui.error "Your score for this failed attempt was: #{log.score}"
             loop do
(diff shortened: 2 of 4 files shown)


lib/vimgolf/cli.rb
```diff
@@ -2,7 +2,6 @@ module VimGolf
 
   GOLFDEBUG    = ENV['GOLFDEBUG'].to_sym rescue false
   GOLFHOST     = ENV['GOLFHOST']     || "http://www.vimgolf.com"
-  GOLFDIFF     = ENV['GOLFDIFF']     || 'diff'
```
igrigorik: Hmm, what about gvim users? If memory serves we had a couple of questions about that before and that's why the option was introduced.. to allow custom diff viewers.

lib/vimgolf/cli.rb
```diff
@@ -61,43 +60,69 @@ def setup
...
+    end
+
+    desc "local INFILE OUTFILE", "launch local challenge"
+    long_desc <<-DESC
+    Launch a local VimGolf challenge. A temporary copy of INFILE is made; the original files will not be touched.
+    DESC
+
+    def local(infile, outfile)
+      # make sure our files are sane
+      if !(File.file?(infile) and File.file?(outfile))
```
igrigorik: should there be a check for infile != outfile?

lib/vimgolf/cli.rb
```diff
@@ -61,43 +60,69 @@ def setup
...
-          diff_files = "\"#{challenge.work_path}\" \"#{challenge.output_path}\""
-          diff = `#{GOLFDIFF} #{diff_files}`
           log = Keylog.new(IO.read(challenge.log_path))
 
           VimGolf.ui.info "\nHere are your keystrokes:"
           VimGolf.ui.print_log log
 
           # Handle incorrect solutions
-          if diff.size > 0
+          if !challenge.correct()
```
igrigorik: Pedantic, but "correct?" would be a better name.

lib/vimgolf/cli.rb
```diff
@@ -177,8 +200,6 @@ def put(id = nil)
       VimGolf.ui.error "If the error persists, please report it to github.com/igrigorik/vimgolf"
     end
 
-    private
```
igrigorik: Why are we exposing other methods as public?

lib/vimgolf/challenge.rb
```diff
@@ -11,7 +11,28 @@ def initialize(id)
       @id = id
     end
 
+    def local(infile, outfile)
+      @remote = false
```
igrigorik: What's the point of @remote? I see it being set, but it's not used anywhere.

lib/vimgolf/cli.rb
```diff
@@ -61,43 +60,69 @@ def setup
...
+    DESC
+
+    def local(infile, outfile)
+      # make sure our files are sane
+      if !(File.file?(infile) and File.file?(outfile))
+        VimGolf.ui.error "INFILE and OUTFILE must exist and be regular files."
+        exit 1
+      end
+
+      challenge = Challenge.new(infile) # use the filename as id
```
igrigorik: in which case we don't need @infile, and don't need to pass it to .local()... right?

lib/vimgolf/challenge.rb
urticadioica: It's only used in the menu in play(), so local challenges don't show an option to upload the solution.

lib/vimgolf/cli.rb
urticadioica: Are you thinking of GOLFSHOWDIFF on the next line?

lib/vimgolf/cli.rb
urticadioica: I thought of that before I commited. If you `vimgolf local foo.txt foo.txt` it behaves as if it's running on separate files that are identical. It's not a bad behavior, so I did nothing about it.

lib/vimgolf/cli.rb
urticadioica: If I were making this with just local play and no remote, I wouldn't even have an id.

lib/vimgolf/cli.rb
urticadioica: I did nothing to the `private` statement, but I had to move most of the old `.put()` into `.play()`, which had to be below `private`. (Thor got mad at me.) Diff is confused. There's a `+ private` higher.

lib/vimgolf/challenge.rb
```diff
@@ -34,9 +55,14 @@ def download
         @data['in']['data'].gsub!(/\r\n/, "\n")
         @data['out']['data'].gsub!(/\r\n/, "\n")
 
-        @type = @data['in']['type']
+        # be sure to sanitize the types
+        @type = @data['in']['type'].gsub(/[^\w-]/, '.')
```
igrigorik: Should we be using shellescape instead? http://www.ruby-doc.org/stdlib-1.9.3/libdoc/shellwords/rdoc/Shellwords.html#method-c-shellescape

lib/vimgolf/cli.rb
igrigorik: See https://github.com/igrigorik/vimgolf/issues/125

lib/vimgolf/cli.rb
igrigorik: I mean.. challenge.id already contains the ID, so passing infile to local() seems unnecessary, since you have the value initialized from the constructor.

lib/vimgolf/challenge.rb
urticadioica: `shellescape` is not portable. The main purpose of this `gsub` is actually to eliminate directory separators, which `shellescape` doesn't do. In fact, it adds a bunch of backslashes, which is a directory separator in Windows.

lib/vimgolf/cli.rb
urticadioica: By "diff viewers" I really think you mean `GOLFSHOWDIFF`. The function of `GOLFDIFF` was to name an external `diff` command to tell you if your solution is right. I replaced that with the `challenge.correct?` method, so `GOLFDIFF` isn't needed any more.

lib/vimgolf/cli.rb
urticadioica: `challenge.new` in this case is just telling the instance what it should call itself. I could just give all local challenges a dummy name:

    challenge = Challenge.new('Local challenge')
    challenge.local(infile, outfile)

This whole branch is pretty minimal. A full refactor would probably put local and remote challenges in separate classes, inheriting the common methods.

lib/vimgolf/cli.rb
igrigorik: Gotcha, thanks.

lib/vimgolf/cli.rb
igrigorik: Fair enough, we can tackle that later.

igrigorik:
great stuff :+1: and thanks for your patience :)

urticadioica:
OK, good. I'd suggest a release anyway so other players can use this stuff. But this branch fixed some serious security problems, so it would be a good time. (I probably should have told you privately. It's pretty bad.)

After it's released, you can get away with some changes:

* Remove the vimrc field from the json download.
* At the bottom of http://www.vimgolf.com/about remove the mention of Windows users needing a diff program.
* Update the keylog.rb on the server. (Is that automatic? Has it already been done? I don't know. But I know #100 still isn't fixed.)
  * You mentioned performance problems from keylog.rb once. Makes sense if you're running it dynamically. I think it can be improved by a constant factor (2-4x), so you'd eventually need to cache results anyway (site isn't shrinking), but I can do that if you still want.
  * I do know optimizing keylog.rb is way easier if you drop Ruby 1.8. It hasn't been supported since 2013. Don't know if we should drop it.

igrigorik:
- "Update the keylog.rb on the server": the site uses the vimgolf gem under the hood, so any changes in client code get picked up with each release.
- Security: yes, good catch. As another TODO: migrate to https!
- Dropping 1.8: not sure what the current version usage is =/

igrigorik:
0.4.7 is live -- thanks for your help with this one!

Issue #147, opened 2015-02-20
Windows CTRL-Z Cheat


dsewtz:
I joined Twitter just to play vimgolf and inadvertently cheated in one of my first submissions:
- After I had worked through the ranks to discover "ZZ" instead of :wq
- I accidentally pressed <CTRL-ZZ> instead of <SHIFT-ZZ>
- and expected the result to be worse than before
- instead it was better by a count of 2 

-> On Windows 7 in cmd or MSYS <CTRL-ZZ> <SHIFT-ZZ> removes the last/next two keystrokes from the count (Here are your keytrokes: "...<ESC>" , line ends with ESC, ZZ is missing).

At first I thought that I had an old gem, because you mention a windows bugfix with 0.4.8, but these are my versions:
ruby: 2.1.5p273   gem:2.2.2   vimgolf:0.4.8

urticadioica:
Let's see if I understand. You try to enter a solution like `ihello<Esc>ZZ`. You type `<C-Z>` instead of `Z` on accident. What happens next? Does vim get suspended? Or do you continue to type `ZZ` afterward?

If you run `vim -Z` from the command line and hit `<C-Z>`, do you get a message like "E145: Shell commands not allowed in rvim"?

dsewtz:
Okay, i did a little more testing:
- I see the message "E145:..." when i hit `<C-Z>`
- After `<C-Z>` no more keystrokes are recorded

- `<C-Z>ihello<ESC>ZZ` becomes `` (nothing) with a count of 0
If i try to upload a 0-score, I get a message "You are not cheating, are you?"

- `ihello<ESC><C-Z><C-Z>ZZ` becomes `ihello<ESC>` with a count of 7
I tried to upload such a solution for "Just the middle" and it worked. I'm at the top with 6 edits now, please delete that entry, if you find the time :-/ SORRY!
My actual keytrokes where `2D3j3D<C-Z>ZZ`
Here is the output:
>Launching VimGolf session for challenge: 54862fbb3f90ac0002904cf5
>
>Here are your keystrokes:
>2D3j3D
>
>Success! Your output matches. Your score: 6
>[w] Upload result and retry the challenge
>[x] Upload result and quit
>[r] Do not upload result and retry the challenge
>[q] Do not upload result and quit
>Choice> x
>Uploading to VimGolf...
>Uploaded entry.
>View the leaderboard: http://www.vimgolf.com/challenges/54862fbb3f90ac0002904cf5
>
[...]

PR #148, opened 2015-02-21
Open keylog files in binary mode

urticadioica:
#147 appears to be caused by opening log files in text mode. In Windows, when you read a file without binary mode and it contains a `^Z`, it counts as an EOF marker. Only text before a `<C-Z>` stroke was counted and reported to the server.

However, in Ruby >=1.9, opening a file in binary mode automatically sets the encoding to `ASCII-8BIT`. That's probably what it should be anyway, so I can just change the marker in keylog.rb. Problem is, this will probably not work on the server, unless the encoding of the log text is switched there too. Otherwise we'll see #100 again. I could also just change the encoding client-side (with some conditionals to keep Ruby 1.8 from breaking). Whatever's more convenient.

### DIFF
--- a/lib/vimgolf/challenge.rb
+++ b/lib/vimgolf/challenge.rb
@@ -91,7 +91,7 @@ module VimGolf
 
         proxy.start(url.host, url.port) do |http|
           request = Net::HTTP::Post.new(url.request_uri)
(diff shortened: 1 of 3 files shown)


urticadioica:
This PR got left behind because of encoding issues (tough to handle when Ruby 1.8 was supported). The relevant changes are all moved to #188, which improves the way encodings are handled.

PR #179, opened 2016-12-30
Make spec suite green

zoso10:
Few things to get the suite green:
 - add `shoulda-matchers` gem to add in missing validation matchers
 - add infer spec type from file location to spec helper
  - specs like controller specs have special local variables to manipulate state; this config will allow us to use that without having to explicitly set the type
 - create custom matcher for `have_fields` expectations for mongo models
  - in lieu of adding another gem ie `mongo-rspec` we can define the custom rspec matcher ourselves since it's a fairly simple one
 - add `database_cleaner` gem to create state for specs
   - there's a bug in edge `database_cleaner` for our version of mongo so we have to lock the version to 1.5.1
 - build state for certain specs
   - some specs were relying on others to run first to create the state they needed to run successfully
 - other misc changes from previous version of rspec


spec/support/mongoid.rb
```diff
@@ -1,3 +1,39 @@
+module Mongoid
+  module Matchers
+    class HaveFieldMatcher
```
igrigorik: Doh, did this get deprecated in a later release?

zoso10: i'm guessing, not entirely sure. I poked around the git history of that project a little but couldn't find anything. do you have a preference for another way of doing it?

nelsonsar:
WOW! Awesome, @zoso10! 👏

spec/support/mongoid.rb
igrigorik: Just sanity checking. This will do.

igrigorik:
👍

Issue #180, opened 2016-12-30
Run CLI tests as part of Travis build


igrigorik:
We have CLI tests in lib/vimgolf/specs, which we should run as part of the Travis build...

igrigorik:
Closing, fixed via https://github.com/igrigorik/vimgolf/issues/181.

PR #181, opened 2016-12-30
Move CLI specs to spec folder

nelsonsar:
This commit moves CLI related specs to spec folder to make it run as part of TravisCI build. This also removes the `helper.rb` file from lib in order to use `spec_helper.rb` in specs.

Related to #180


igrigorik:
Awesome work. Thanks!

PR #185, opened 2017-01-03
Update rspec syntax from should(rspec 2) to expect(rspec 3)

hvenables:
Vimgolf was using the old syntax of rspec. I thought i would quickly update this before getting started on extending the code coverage

PR #186, opened 2017-01-04
Add tests to models

nelsonsar:
# Main changes

## Add factories to help test scenario creation

This commit adds FactoryGirl and Faker and create factories for: `Challenge`, `User` and `Entry` models. The idea is been able to create test scenarios in a easier way.

## Refactoring on Entry model

`Entry` model has two new methods:

* `any_owned_by`: Checks if a user has at least one entry in the given criteria (most of the time it's scoped by Challenge model) and return `true` or `false`.

* `top_by_user`: Returns an array containing user's most recent entries with best score.

## General changes

* Changed old Hash syntax for the new one.
* Add tests to all models.

P.S.: Could not test `count_entries` because of MongoDB version. Which version should I install?

Any suggestion about tests, syntax or method names are very welcome :)


spec/factories/entry.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
```
zoso10: i'm a fan of keeping all factories in a single file and keeping the file ordered alphabetically. especially for smaller data models such as this project's it shouldn't be too bad to navigate. i'd love to hear other's thoughts

spec/models/challenge_spec.rb
```diff
@@ -1,49 +1,103 @@
...
-    p u.save
-
-    p u
-    expect(u.challenges.size).to eq(1)
-    expect(u.challenges.first).to eq(c)
-
-    u = User.find(u.id)
-    p u
-    expect(u.challenges.size).to eq(1)
+describe Challenge, type: :model do
```
zoso10: we infer spec type automatically so we shouldn't have to specify type :)

spec/support/factory_girl.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.configure do |config|
+  config.include FactoryGirl::Syntax::Methods
```
zoso10: thoughts on adding this configuration directly in `spec_helper.rb`? that seems to be the convention we're leaning towards

hvenables: Yeah, I would agree

Gemfile
```diff
@@ -25,7 +25,7 @@ group :test, :development do
 	  gem lib, git: "git://github.com/rspec/#{lib}.git"
 	end
   gem "shoulda-matchers"
-  gem "database_cleaner", "1.5.3"
```
zoso10: sadly, there's a bug in 1.5.3 that won't work with our version of mongo 😞 I opened a pr in the project but it hasn't been super active so I'm not sure if it'll make it in there

spec/factories/entry.rb
nelsonsar: Right now it makes more sense. We have just three of them and they are very, very small. I'll change it :)

spec/models/challenge_spec.rb
nelsonsar: Okdok!

Gemfile
nelsonsar: Yep, I changed it in my last commit :)

igrigorik:
> P.S.: Could not test count_entries because of MongoDB version. Which version should I install?

We're currently on 2.6.5, and we should probably update.. :)

spec/models/challenge_spec.rb
```diff
@@ -2,48 +2,102 @@
...
-    expect(u.challenges.size).to eq(1)
+  describe 'Validations' do
+    it { should have_fields(:title, :description, :input, :output, :diff) }
 
+    it { should validate_presence_of(:title) }
+    it { should validate_presence_of(:description) }
+
+    it { should validate_length_of(:input) }
+    it { should validate_length_of(:output) }
+    it { should validate_length_of(:diff) }
```
hvenables: We recently updated the rspec syntax to get rid of deprecated should, I think that this should be  `is_expected.to` [#185](https://github.com/igrigorik/vimgolf/pull/185)

spec/models/comment_spec.rb
```diff
@@ -1,7 +1,31 @@
...
+
+    context 'comment with comment and nickname' do
+      it 'return true' do
+        comment.comment = 'Foo'
+        comment.nickname = 'Bar'
+
+        expect(comment).to be_valid
+      end
+    end
+  end
```
hvenables: I am not sure what this is testing that we didnt already have in place before? I may have missed something tho? :)

zoso10: it varies between people on the best strategy to test validations. some people prefer using `shoulda_matchers` that provides the `is_expected_to validate_presence_of(:foo)` syntax and others prefer using the predicate matcher with `valid?` using real examples.

hvenables: Guess we just need to decide on what the convention will be? Personally i like `shoulda_matchers` think they give a really nice concise syntax

spec/factories.rb
```diff
@@ -0,0 +1,26 @@
...
+    description :test
+    input :a
+    output :b
+    diff :c
+    user
+  end
+
+  factory :entry do
+    script :a
+    created_at Time.now
```
zoso10: small nit; when we create `Entry`s they're in utc. shouldn't matter for the factory here, but might be nice to keep it consistent

igrigorik:
Are we still on track to land this one?

zoso10:
@nelsonsar if u want to fix the `Gemfile` merge conflict I'll go ahead and pull this in

igrigorik:
Friendly bump.. :-)

@nelsonsar @zoso10 can we pull this one over the finish line?

zoso10:
merged #197 which contained the commits in this PR and fixed the merge conflicts. closing

PR #188, opened 2017-01-07
The keycodes have changed again :(

urticadioica:
I've noticed a few keycode issues lately, and figured Bram probably messed them up again. Sure enough, [he did](https://github.com/vim/vim/commit/85b11769ab507c7df93f319fd964fa579701b76b#diff-7e3a7ac200609674213959e36dfb4d3d). The affected keycodes, especially more popular ones from `<C-Left>` to `<C-End>`, are hopelessly wrong. There's nothing we can do to fix them.

The options are:
* Use the old values (keep them as they are).
* Use the new values.
* Replace the keycodes with a joke. Because that's what they are at this point.

### DIFF
--- a/lib/vimgolf/lib/vimgolf/keylog.rb
+++ b/lib/vimgolf/lib/vimgolf/keylog.rb
@@ -170,6 +170,10 @@ module VimGolf
               when "\xfd\x35"; nil # KE_IGNORE
               #when "\xfd\x36"; "KE_TAB"
               #when "\xfd\x37"; "KE_S_TAB_OLD"
+
+              # Vim 7.4.1433 removed KE_SNIFF. Unfortunately, this changed the
+              # offset of every keycode after it. No keycode after this point
+              # can be trusted.
               #when "\xfd\x38"; "KE_SNIFF"
               #when "\xfd\x39"; "KE_XF1"
               #when "\xfd\x3a"; "KE_XF2"
@@ -223,7 +227,7 @@ module VimGolf
               #when "\xfd\x5f"; "KE_DROP"
               #when "\xfd\x60"; "KE_CURSORHOLD"
               when "\xfd\x60"; nil # 7.2 Focus Gained compat
-              #when "\xfd\x61"; "KE_NOP"
+              when "\xfd\x61"; nil # Focus Gained (GVIM) (>7.4.1433)
               when "\xfd\x62"; nil # Focus Gained (GVIM)
               when "\xfd\x63"; nil # Focus Lost (GVIM)
 
--- a/lib/vimgolf/lib/vimgolf/keylog.rb
+++ b/lib/vimgolf/lib/vimgolf/keylog.rb
@@ -1,4 +1,5 @@
-# encoding: UTF-8
+# encoding: ASCII-8BIT
+# Force encoding of string literals. Must match solution text.
 
 module VimGolf
   class Keylog
@@ -8,7 +9,8 @@ module VimGolf
     alias_method :score   , :count
 
     def initialize(input)
-      @input = input
+      # Force encoding of solution text. Must match string literals.
+      @input = input.b
     end
 
     def to_s(sep = '')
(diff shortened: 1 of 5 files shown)


igrigorik:
That's a tricky one.. Ideally, I guess, we'd want to maintain correct keycodes for existing solutions, but use the new offsets moving forward?

We do have the timestamps on each submission... If we're willing to go the extra mile, in theory we could add conditional logic based on timestamp of the submission; have the parser take an optional timestamp and apply the appropriate logic? =/

urticadioica:
> If we're willing to go the extra mile, in theory we could add conditional logic based on timestamp of the submission; have the parser take an optional timestamp and apply the appropriate logic? =/

That's actually kinda doable... Basically ditch the `case` for the `KS_EXTRA`s, and select the keystroke from a list, indexed by the 3rd byte (like a reverse of the C enum that makes this mess in the first place). Set the base list to the current keymap.h values, and insert `KE_SNIFF` to the right spot if date corresponds before 7.4.1433, and remove `KE_MOUSELEFT` and `KE_MOUSERIGHT` if before date corresponding to 7.3c. (That's the other time this happened before, see [this comment](https://github.com/urticadioica/vimgolf/blob/6ed7b3d4069b185923943e9b10b6561ca3f994df/lib/vimgolf/lib/vimgolf/keylog.rb#L199).)

* It wouldn't be 100% right. It would be interesting to look at a database dump of solutions with `KS_EXTRA` and see when the users tend to start adopting the new versions after the two changes to-date. Or we could just guess based on release dates. ;P
* I would say it might make the parsing slower, but nothing on Earth could be slower than what we've got now. I've tried refactoring before, but keeping Ruby 1.8 support client-side is *extremely* annoying. If I ever touch this file again in a serious way again, I'm killing 1.8 support.

igrigorik:
Ruby 1.8 has been [deprecated for 5+ years](https://www.ruby-lang.org/en/news/2011/10/06/plans-for-1-8-7/). In fact, same [1.9.3 for ~2 years](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/). Which is to say, I'm 100% for requiring modern Ruby (2.x+).

> It wouldn't be 100% right. It would be interesting to look at a database dump of solutions with KS_EXTRA and see when the users tend to start adopting the new versions after the two changes to-date. Or we could just guess based on release dates. ;P

Ah right, hmm. Could we auto-detect this based on the keylog? =/ Also, we could start logging the VIM version in uploads, if that makes our life easier moving forward?

urticadioica:
Got carried away and refactored everything. Keylog is 2.5x faster now, so that's nice.

> Also, we could start logging the VIM version in uploads, if that makes our life easier moving forward?

A variation I thought about was checking the version CLI-side, and modifying the keylog before uploading to something standard. Then I thought about how I'd figure out which Vim version we had. Do I parse `vim --version` and check what patches are applied? Some distros skip certain patches, and some probably mess with the `--version` output. Getting it basically right is easy enough, but getting it exactly right is terrifying.

Or I could write a Vim script that opens the log file. Vim has better ways to check patch level than parsing a string. But you have to call Vim an extra time every session, and you've got to make the script work in whatever Vim the user provided. What if they messed with `GOLFVIM`? What if they asked for `gvim` and now I'm popping an extra window in their face? There's a lot I'd have to get right, for pretty small benefit.

> Could we auto-detect this based on the keylog?

It's possible. If I read a solution with human eyes, I can figure out if there's something wrong. The "easiest" way is to run the log through Vim again and check if it works (client-side or server-side). Doable, but given the difficulty, the benefit is questionable.

I did some research on solutions site-wide. I found 152 solutions with a keystroke that could be affected by this problem. 10 of those were displayed wrong due to the 7.3c change ("early"), and 6 were affected by the recent 7.4.1433 change ("late"). 16 mistakes out of about 59,000 total solutions, so that's the scope of the problem.

I wanted to know if checking the date would be useful. For "early", the best you can do is to fix 3 wrongly-displayed solutions, while breaking 1 currently correct. So I'll call that totally useless. The first "late" solution was submitted 2016-04-18, and since then no "middle" solution using the affected keystrokes has been submitted. So 6 to 0, and (very slowly) growing. A date solution isn't hard, but it's not a tragedy if no extra action is taken. I did want to add the comments in the first commit to at least document it. I can do more if necessary.

lib/vimgolf/lib/vimgolf/keylog.rb
```diff
@@ -1,258 +1,267 @@
...
+    (32..126).each {|c| @@kc_1byte[c] = c.chr } # Printing chars
+    @@kc_1byte[0x1b] = "<Esc>" # Special names for a few control chars
+    @@kc_1byte[0x0d] = "<CR>"
+    @@kc_1byte[0x0a] = "<NL>"
+    @@kc_1byte[0x09] = "<Tab>"
 
-              when "kB"; "<S-Tab>"
-              when "\xffX"; "<C-@>"
+    # After this date, assume KE_SNIFF is removed
+    @@sniff_date = Time.utc(2016, 4)
```
igrigorik: Any reason why we can't make this a const? 

Ditto for other @@'s.. we can create and freeze the hashes? If we never change any of these, perhaps that's a better route?

igrigorik:
> Got carried away and refactored everything. Keylog is 2.5x faster now, so that's nice.

You, sir, are my hero.

> A variation I thought about was checking the version CLI-side, and modifying the keylog before uploading to something standard.

A normalized format _would_ be nice. However, my experience shows that whenever you're attempting such a thing, you should also _always_ log the original, such that you can reprocess and fix errors after the fact. Otherwise, your normalize out both the good and the bad. Which is to say, we _could_ try to normalize and score based on that, but we should log the original data regardless.

That aside.. The update looks sane, modulo question about freezing vs class variables. Once we resolve that, I think we're good to land this. Also, kudos for adding additional tests!

urticadioica:
You're right on the constants. A little cleaner, and benchmarks 15-20% faster too. The name lookups on class variables must be slow...

I figured I'd have to change a few things once I could test Ruby 1.9. I don't have one on my box, but I don't want to break it until I have to, so CI is nice for that.

igrigorik:
Nice, thanks for the quick turn around. It looks like it's ready to go? If so, I can squash and merge.

urticadioica:
I messed up the history a bit there, but yeah, I'm good.

igrigorik:
Kudos!

zuloloxi:
I could try to create un issue on vim, for the affected keycodes, especially more popular ones from  <C-Left>  to  <C-End> , that are  wrong, but must explain what is happening with vim, and must have un example for testing.

zuloloxi:
The simple solution for **&lt;C-End**&gt; display **&lt;C-Home**&gt; is to rollback the line 214 in src/keymap.h before vim 7.4.1433, patch [Problem:  The Sniff interface is no longer useful, the tool has not been available for may years, and Solution:   Delete the Sniff interface and related code](https://github.com/vim/vim/commit/85b11769ab507c7df93f319fd964fa579701b76b#diff-7e3a7ac200609674213959e36dfb4d3d), like:
```
      , KE_TAB		/* unshifted TAB key */
      , KE_S_TAB_OLD	/* shifted TAB key (no longer used) */

      , KE_SNIFF		/* SNiFF+ input waiting (no longer used) re-added this line - 214 */

      , KE_XF1		/* extra vt100 function keys for xterm */
      , KE_XF2
      , KE_XF3
...
```

Tested and works with vim 8.0.647, patched KE_SNIFF	!!!

igrigorik:
I'll defer to @urticadioica on this one.. :-)

zuloloxi:
Created issue 1810 *A few keycode issues* on Vim_dev https://groups.google.com/forum/#!topic/vim_dev/1R4YkuuQruI

urticadioica:
Uh-huh.

They really shouldn't fix this. If they did fix it, that would be just another bug for us. It would be better if they hadn't done this patch in the first place, but now that they've done it, it would be a problem for everyone if they rolled it back.

zuloloxi:
I described the problem **&lt;C-Home&gt;** instead of **&lt;C-End&gt;** detected in the challenge [Missing Library and a Typo](http://www.vimgolf.com/challenges/59553bd164628d0009000038), that I don't have with vim 7.4.1432, and with the patch above.

zuloloxi:
I create a zulolosi solution at before last solution, with 43 keystrokes, in the challenge [Missing Library and a Typo](http://www.vimgolf.com/challenges/59553bd164628d0009000038), : 
 Used this keystrokes in order :
```
<C-End>xP<C-Home><C-Right><C-Left>wwwwwwwwwwwwwwwwwwwwwwwwwwYpwwwRa<C-P><Esc>ZZ 
```
 but it display in the web text solution :
 ```
<C-Home>xP<C-Right><C-Left><C-Right>wwwwwwwwwwwwwwwwwwwwwwwwwwYpwwwRa<C-P><Esc>ZZ 
```
 The keys codes displayed by the vimgolf client:
``` 
<fd-57>xP<fd-56><fd-55><fd-54>wwwwwwwwwwwwwwwwwwwwwwwwwwYpwwwRa<C-P><Esc>ZZ 
```
I noticed that:
```
<C-Home> and <C-Left> keystrokes display the same text : <C-Right> in the web solution, 
but the keys codes of the vimgolf client are different for:
 <C-Home> is <fd-56> 
        and for 
 <C-Left> is <fd-54>
```

I tested a solution with vim 7.4.1432, the keycodes of vimgolf client are:
```
<fd-58>xP<fd-57><fd-56><fd-55>wwwwwwwwwwwwwwwwwwwwwwYpfsRa<C-P><Esc>ZZ
```
and the web text solution  is correct, same as the keystrokes used  :
```
<C-End>xP<C-Home><C-Right><C-Left>wwwwwwwwwwwwwwwwwwwwwwYpfsRa<C-P><Esc>ZZ 
```
May be, for developing the vimgolf client,  it was used the vim internal keycodes, in [keylog.rb](https://github.com/igrigorik/vimgolf/blob/9ad95570a96cd6f9f5d764d80d2e55c621a26347/lib/vimgolf/lib/vimgolf/keylog.rb#L244), Mr. @brammool [commented here at 3 juil. 2017 22:56](https://github.com/vim/vim/issues/1810#issuecomment-312732250)

**Comment by urticadioica** [comment]
@zuloloxi THIS IS NOT A BUG IN VIM. If the Vim devs did anything about this, it would only cause more trouble for us. It would be great if you dropped this.

**Comment by zuloloxi** [comment]
ok ! I agree ! you must take a look of [disscution](https://github.com/vim/vim/issues/1810),  may be you could contribute, and you must change [keylog.rb](https://github.com/igrigorik/vimgolf/blob/9ad95570a96cd6f9f5d764d80d2e55c621a26347/lib/vimgolf/lib/vimgolf/keylog.rb#L244) to use the new keycodes.

**Comment by zuloloxi** [comment]
The long list of keycodes that changed and decremented (-1)  after vim 7.4.1432 : 
```
- , KE_SNIFF		/* SNiFF+ input waiting */ deleted line in vim 7.4.1432

    , KE_XF1		/* extra vt100 function keys for xterm */
    , KE_XF2
    , KE_XF3
    , KE_XF4
    , KE_XEND		/* extra (vt100) end key for xterm */
    , KE_ZEND		/* extra (vt100) end key for xterm */
    , KE_XHOME		/* extra (vt100) home key for xterm */
    , KE_ZHOME		/* extra (vt100) home key for xterm */
    , KE_XUP		/* extra vt100 cursor keys for xterm */
    , KE_XDOWN
    , KE_XLEFT
    , KE_XRIGHT

    , KE_LEFTMOUSE_NM	/* non-mappable Left mouse button click */
    , KE_LEFTRELEASE_NM	/* non-mappable left mouse button release */

    , KE_S_XF1		/* extra vt100 shifted function keys for xterm */
    , KE_S_XF2
    , KE_S_XF3
    , KE_S_XF4

    /* NOTE: The scroll wheel events are inverted: i.e. UP is the same as
     * moving the actual scroll wheel down, LEFT is the same as moving the
     * scroll wheel right. */
    , KE_MOUSEDOWN	/* scroll wheel pseudo-button Down */
    , KE_MOUSEUP	/* scroll wheel pseudo-button Up */
    , KE_MOUSELEFT	/* scroll wheel pseudo-button Left */
    , KE_MOUSERIGHT	/* scroll wheel pseudo-button Right */

    , KE_KINS		/* keypad Insert key */
    , KE_KDEL		/* keypad Delete key */

    , KE_CSI		/* CSI typed directly */
    , KE_SNR		/* <SNR> */
    , KE_PLUG		/* <Plug> */
    , KE_CMDWIN		/* open command-line window from Command-line Mode */

    , KE_C_LEFT		/* control-left */
    , KE_C_RIGHT	/* control-right */
    , KE_C_HOME		/* control-home */
    , KE_C_END		/* control-end */

    , KE_X1MOUSE	/* X1/X2 mouse-buttons */
    , KE_X1DRAG
    , KE_X1RELEASE
    , KE_X2MOUSE
    , KE_X2DRAG
    , KE_X2RELEASE

    , KE_DROP		/* DnD data is available */
    , KE_CURSORHOLD	/* CursorHold event */
    , KE_NOP		/* doesn't do something */
    , KE_FOCUSGAINED	/* focus gained */
    , KE_FOCUSLOST	/* focus lost */
```
and  sure enough display a wrong text in *vimgolf* **Solutions** !

zuloloxi:
Sorry, I dropped, but Mr. Vim Boss @brammool and other developers , decided to re-work the KE_xxxx and to use fixed numbers for the **key_extra** enum,  in :[patch 8.0.0697: recorded key sequences may become invalid](https://github.com/vim/vim/commit/8858498516108432453526f07783f14c9196e112#diff-7e3a7ac200609674213959e36dfb4d3dR217), sorry if it cause some trouble in vimgolf. For now I see that the text in solutions in vimgolf re-work again ...

PR #191, opened 2017-01-13
Add feature specs for better integration coverage of web app

zoso10:
i decided to target one of the most important endpoints first which is the `/entry` endpoint that the gem uses to submit challenge entries. i noticed it was basically a json endpoint so it's best not to use capybara to test it. I used a request spec which is most common for api endpoints.

_note_: ~i'll be breaking up the feature specs into a number of PRs to make them more digestible and easier to review.~ they're broken up into digestible commits. i decided i didn't want to have to deal with rebasing, _but_ upon request i can close this one and make separate PRs if people wish. also, once #186 is merged ill change this to use factories.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -9,6 +9,7 @@ require 'rspec/rails'
 # Requires supporting ruby files with custom matchers and macros, etc,
 # in spec/support/ and its subdirectories.
 Dir[Rails.root.join("spec/support/**/*.rb")].each {|f| require f}
+Capybara.default_driver = :rack_test
 
 RSpec.configure do |config|
   config.mock_with :rspec
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -22,6 +22,10 @@ RSpec.configure do |config|
     DatabaseCleaner.clean_with(:truncation)
   end
 
+  config.before(:each) do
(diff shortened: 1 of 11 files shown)


spec/spec_helper.rb
```diff
@@ -9,6 +9,7 @@
 # Requires supporting ruby files with custom matchers and macros, etc,
 # in spec/support/ and its subdirectories.
 Dir[Rails.root.join("spec/support/**/*.rb")].each {|f| require f}
+Capybara.default_driver = :rack_test
```
hvenables: Sure rack test is fine for now, if we ever need a driver that can execute javascript poltergeist is my preferred driver

zoso10: agreed, i always have bad luck with webkit and getting qt to play nicely together and selenium seems heavyweight.
there's really not a ton of js in the app right now anyway, the twitter feed integration is the biggest thing. we could probably test that separately anyway.

zoso10:
@hvenables there's a few more specs id like to add, but i think this covers critical paths. i feel a lot more comfortable moving on upgrading, thoughts?

hvenables:
@zoso10 Agreed, Im more comfortable but i would prefer the coverage to be a little higher before starting on upgrading.

igrigorik:
p.s... awesome work guys!

PR #193, opened 2017-01-16
Test just cli, or web AND cli

urticadioica:
It's useful to test the cli without needing to install rails or mongodb.
You should still be able to run all the tests from the project root as
before, but now running `rake` or `rspec` from the CLI root runs just those
tests, without extra dependencies.

Also tries to make Travis test multiple versions of Ruby. The server
needs a particular version, but the CLI is expected to run on many
versions. On the non-server versions, Travis should only run the CLI
tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,21 @@
 language: ruby
-rvm:
-    - 2.1.5
+matrix:
+  include:
+    - rvm: 2.1.5 # Test web AND cli with specific version
+      gemfile: Gemfile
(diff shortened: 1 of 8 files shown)


igrigorik:
👍 

Any other updates, or is this good to merge?

urticadioica:
Should be good. (Took me a few tries testing the Travis config, since it's awkward to do locally, but I've done what I want for now.)

igrigorik:
Great stuff -- thanks!

PR #194, opened 2017-01-20
Feature/entry feature spec

hvenables:
Feature spec for Entries. Covering the other two routes in this controller. Had to add poltergiest as both the commenting and deleteing require you to click on Commen / Edit which uses javascript to then reveal a form.

### DIFF
--- a/app/controllers/entry_controller.rb
+++ b/app/controllers/entry_controller.rb
@@ -8,7 +8,7 @@ class EntryController < ApplicationController
       @entry.comments.push Comment.new(:comment => params[:comment][:text], :nickname => current_user.nickname)
       @challenge.save
     end
-  rescue Exception
+  rescue
 
   ensure
     redirect_to challenge_path(params[:challenge])
@@ -55,7 +55,7 @@ class EntryController < ApplicationController
     end
 
     redirect_to challenge_path(params[:challenge])
-  rescue Exception
+  rescue
     redirect_to root_path
   end
 
@@ -65,7 +65,7 @@ class EntryController < ApplicationController
       @challenge = Challenge.find(params[:challenge])
       @entry = @challenge.entries.find(params[:entry])
 
-    rescue Exception
+    rescue
       respond_to do |format|
         format.json {
           render :json => {'status' => 'failed'}, :status => 400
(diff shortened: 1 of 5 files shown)


app/controllers/entry_controller.rb
```diff
@@ -4,16 +4,14 @@ class EntryController < ApplicationController
...
-      end
-
-    rescue Exception => e
-    ensure
-      redirect_to challenge_path(params[:challenge])
+    if @challenge.participator?(current_user) && params[:comment] && params[:comment][:text].size > 0
+      @entry.comments.push Comment.new(:comment => params[:comment][:text], :nickname => current_user.nickname)
+      @challenge.save
     end
+  rescue Exception
```
hvenables: Throughout this controller we are rescuing Exception, not sure this is the best strategy. If we are going to use such a large scope to rescue would it not be better to rescue StandardError?

igrigorik: Yeah, I agree, that was overly eager on my part.. We can scope it down.

app/controllers/entry_controller.rb
```diff
@@ -4,16 +4,14 @@ class EntryController < ApplicationController
...
-      end
-
-    rescue Exception => e
-    ensure
-      redirect_to challenge_path(params[:challenge])
+    if @challenge.participator?(current_user) && params[:comment] && params[:comment][:text].size > 0
+      @entry.comments.push Comment.new(:comment => params[:comment][:text], :nickname => current_user.nickname)
+      @challenge.save
     end
+  rescue
```
zoso10: this feels a little odd. could `@entry` possibly be `nil` is that why this is here? not a huge deal but curious what your thoughts on adding the redirect here as well? it's usually bad practice to have empty `rescue` blocks

zoso10:
willing to defer on the codeclimate issue on suppressing exceptions for later so we can continue. thoughts @igrigorik ?

igrigorik:
sgtm.

hvenables:
Sorry guys was a little busy last week or so... Ill sort out that rescue today and make a very minuscule pull request

zoso10:
@hvenables totally understandable, no worries!

PR #195, opened 2017-01-30
Removed empty rescue blocks in controllers

hvenables:
Rather than having empty rescue block we could have ensures and returns. Its not great but im not sure what we are rescuing against here? I think we could remove the ensure and just guard against `@entry` or `@challenge` being nil? what do you think?

### DIFF
--- a/app/controllers/entry_controller.rb
+++ b/app/controllers/entry_controller.rb
@@ -1,13 +1,14 @@
 class EntryController < ApplicationController
 
   before_filter :login, :only => [:comment]
-  before_filter :load_entry, :only => [:comment]
+  before_filter :load_entry, :only => [:comment, :destroy]
 
   def comment
-    if @challenge.participator?(current_user) && @entry && params[:comment].try(:[], :text).try(:present?)
+    if @challenge.participator?(current_user) && @entry && params.fetch(:comment, {})[:text].present?
       @entry.comments.push Comment.new(:comment => params[:comment][:text], :nickname => current_user.nickname)
       @challenge.save
     end
+
     redirect_to challenge_path(params[:challenge])
   end
 
@@ -41,33 +42,25 @@ class EntryController < ApplicationController
   end
 
   def destroy
-    challenge = Challenge.find(params[:challenge])
-    entry = challenge.entries.find(params[:entry])
-
-    if challenge.owner?(current_user) || entry.owner?(current_user)
-      entry.destroy
-      challenge.save
-
-      flash[:notice] = "Deleted entry"
+    if @entry && (@challenge.owner?(current_user) || @entry.owner?(current_user))
+      @entry.destroy
+      @challenge.save
     end
 
     redirect_to challenge_path(params[:challenge])
-  rescue
-    redirect_to root_path
   end
 
   private
 
-    def load_entry
-      @challenge = Challenge.find(params[:challenge])
-      @entry = @challenge.entries.find(params[:entry])
-
-    rescue
-      respond_to do |format|
-        format.json {
-          render :json => {'status' => 'failed'}, :status => 400
-        }
-        format.html { redirect_to root_path }
-      end
+  def load_entry
+    @challenge = Challenge.find(params[:challenge])
+    @entry = @challenge.entries.find(params[:entry])
+  rescue
+    respond_to do |format|
+      format.json {
+        render :json => {'status' => 'failed'}, :status => 400
+      }
+      format.html { redirect_to root_path }
     end
+  end
 end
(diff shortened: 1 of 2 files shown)


igrigorik:
code climate: "Do not return from an ensure block"... ;-)

> I think we could remove the ensure and just guard against @entry or @challenge being nil? what do you think?

Yep, sounds reasonable to me.

app/controllers/entry_controller.rb
```diff
@@ -4,13 +4,10 @@ class EntryController < ApplicationController
   before_filter :load_entry, :only => [:comment]
 
   def comment
-    if @challenge.participator?(current_user) && params[:comment] && params[:comment][:text].size > 0
+    if @challenge.participator?(current_user) && @entry && params[:comment].try(:[], :text).try(:present?)
```
zoso10: thoughts on using `fetch` with a default? may make things a little more clear. up to you

hvenables: Yeah its probably a little clearer, ill update it when i get home tonight

app/controllers/entry_controller.rb
```diff
@@ -1,16 +1,14 @@
 class EntryController < ApplicationController
 
   before_filter :login, :only => [:comment]
-  before_filter :load_entry, :only => [:comment]
+  before_filter :load_entry, :only => [:comment, :destroy]
 
   def comment
-    if @challenge.participator?(current_user) && params[:comment] && params[:comment][:text].size > 0
+    if @challenge.participator?(current_user) && @entry && params.fetch(:comment, {})[:text].present?
```
hvenables: @zoso10 Is this what you meant with fetch?

app/controllers/entry_controller.rb
```diff
@@ -44,33 +42,25 @@ def create
...
   def destroy
-    challenge = Challenge.find(params[:challenge])
-    entry = challenge.entries.find(params[:entry])
-
-    if challenge.owner?(current_user) || entry.owner?(current_user)
-      entry.destroy
-      challenge.save
-
-      flash[:notice] = "Deleted entry"
+    if @entry && (@challenge.owner?(current_user) || @entry.owner?(current_user))
```
hvenables: I realised here we are doing what the method load entry is already doing, so i added the callback for destroy method. I also thought that if we guard against `@entry` being nil then i think the rescue is not neccessary? what do you think?

app/controllers/entry_controller.rb
```diff
@@ -44,33 +42,25 @@ def create
...
+      @entry.destroy
+      @challenge.save
     end
 
     redirect_to challenge_path(params[:challenge])
-  rescue
-    redirect_to root_path
   end
 
   private
```
hvenables: This indentation was one level too deep i think

app/controllers/entry_controller.rb
```diff
@@ -44,33 +42,25 @@ def create
...
-      @entry = @challenge.entries.find(params[:entry])
-
-    rescue
-      respond_to do |format|
-        format.json {
-          render :json => {'status' => 'failed'}, :status => 400
-        }
-        format.html { redirect_to root_path }
-      end
+  def load_entry
```
hvenables: Think this was one indentation level to deep

app/controllers/entry_controller.rb
```diff
@@ -44,33 +42,25 @@ def create
...
 
   def destroy
-    challenge = Challenge.find(params[:challenge])
-    entry = challenge.entries.find(params[:entry])
-
-    if challenge.owner?(current_user) || entry.owner?(current_user)
-      entry.destroy
-      challenge.save
-
-      flash[:notice] = "Deleted entry"
```
hvenables: Added that on the last PR but realised we are not using flash messages.

app/controllers/entry_controller.rb
zoso10: if we add nil checks in the `if` branch we'll need to add an `else` branch that does the redirect or renders the json. the before action keeps things a little dry. since we're using `find` in the before action it'll raise and be caught by the rescue and do the redirect for us

app/controllers/entry_controller.rb
zoso10: yup, i think it makes things clearer. others thoughts?

app/controllers/entry_controller.rb
hvenables: Yeah, think that the rescue in the `load_entry` will redirect if their is an error with `find` and as the `redirect_to` is outside the `if` branch we dont need an `else`

igrigorik:
Great work -- thanks guys!

PR #196, opened 2017-03-23
Link to vimgolf.el

timvisher:
I fixed up vimgolf.el and verified that all the functionality I expect to work does. I think it makes sense to maintain this as an entirely separate entity given that we're never going to allow submissions from Emacs to vimgolf the site. I think a link makes perfect sense and probably encourages people to develop other editor interfaces as well! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,13 @@ $> vimgolf setup
 $> vimgolf put [challenge ID]
 ```
 
+# Playing from other editors
+
+## Emacs
+
+There's a lightly maintained interface to play VimGolf challenges in Emacs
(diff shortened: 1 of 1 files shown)


urticadioica:
@timvisher Have you looked at [Anarchy Golf](http://golf.shinh.org)? There's no Emacs option there right now, though there probably should be. I don't know what would be involved in adding one, or whether it's worth trying, but it's worth checking out if you haven't.

timvisher:
@urticadioica Nice! I had not heard of that. Looks fun. :)

timvisher:
@urticadioica This you? https://twitter.com/udioica

igrigorik:
👍  thanks!

PR #197, opened 2017-04-01
Add tests to models

house-owl:
# Main changes

## Add factories to help test scenario creation

This commit adds FactoryGirl and Faker and create factories for: `Challenge`, `User` and `Entry` models. The idea is been able to create test scenarios in a easier way.

## Refactoring on Entry model

`Entry` model has two new methods:

* `any_owned_by`: Checks if a user has at least one entry in the given criteria (most of the time it's scoped by Challenge model) and return `true` or `false`.

* `top_by_user`: Returns an array containing user's most recent entries with best score.

## General changes

Changed old Hash syntax for the new one.
Add tests to all models.
P.S.: Could not test `count_entries` because of MongoDB version. Which version should I install?

Any suggestion about tests, syntax or method names are very welcome :)


house-owl:
Noticed that https://github.com/igrigorik/vimgolf/pull/186 was getting stale and addressed the merge conflicts and suggestions in the original PR. 

@nelsonsar @zoso10 @igrigorik

house-owl:
Plagiarized @nelsonsar's PR description, to preserve its original intent.

igrigorik:
@house-owl nice, thank you for picking this up! 

@nelsonsar @zoso10 do you have bandwidth to review? I'd prefer to defer to you guys on this one :)

spec/factories.rb
```diff
@@ -0,0 +1,26 @@
...
+    description :test
+    input :a
+    output :b
+    diff :c
+    user
+  end
+
+  factory :entry do
+    script :a
+    created_at Time.now.utc
```
zoso10: not sure when `Time.current` was introduced in rails but it does this exact thing. totally fine to leave for now. this is a good find tho, could cause some test inconsistencies :)

zoso10:
awesome, thanks for reviving this! 👍

spec/factories.rb
house-owl: I will make another PR to convert this across the site. Thank you for pointing it out!

igrigorik:
\o/

PR #198, opened 2017-04-13
Add rspec after each block to reset `VimGolf.ui` attribute to imitate initial class load

zoso10:
you can trigger a failure with `bundle exec rspec --seed 15862 spec/lib/cli_spec.rb`.
essentially any time `VimGolf::CLI.start` is called before the `"sets up VimGolf.ui"` spec is run it will fail because the `VimGolf.ui` attribute has already been set and the state persists between specs.
i have a rails upgrade branch that adds in `--order rand` option for rspec which will see intermittent errors until this is pulled in 

also I added a couple reviewers based on github suggestions. hope that's ok :)

### DIFF
--- a/spec/cli_helper.rb
+++ b/spec/cli_helper.rb
@@ -14,3 +14,9 @@ module Kernel
   alias capture_stdout capture_stdio
 end
 
+RSpec.configure do |config|
+  config.after(:each) do
+    # To imitate the initial class load before calling `VimGolf::CLI.start`
+    VimGolf.ui = VimGolf::UI.new
+  end
+end


spec/cli_helper.rb
```diff
@@ -14,3 +14,9 @@ def capture_stdio(input = nil, &block)
   alias capture_stdout capture_stdio
 end
 
+RSpec.configure do |config|
+  config.after(:each) do
+    # To imitate the initial class load before calling `VimGolf::CLI.start`
+    VimGolf.ui = VimGolf::UI.new
```
igrigorik: Hmm, isn't this already handled by https://github.com/igrigorik/vimgolf/blob/master/lib/vimgolf/lib/vimgolf/cli.rb#L31? I think I'm missing something obvious here. :)

zoso10: `ui` is a class attribute and once the class is loaded in the spec helper the class attribute state persists between individual spec runs. so if [this](https://github.com/igrigorik/vimgolf/blob/master/spec/lib/cli_spec.rb#L18) spec runs before [this](https://github.com/igrigorik/vimgolf/blob/master/spec/lib/cli_spec.rb#L8) one `VimGolf.ui` will still be an instance of `VimGolf::UI` from the first spec calling `start`. we have to artificially reset the state after each run

PR #199, opened 2017-04-13
Upgrade to Rails 3.1.0

zoso10:
Sorry this took a lot longer than expected 😅 

(mostly) normal upgrade here:
 - bumped the rails gem and any necessary gems
 - ran the rake upgrade task to have more modern config file formatting
 - I commented out the gems for asset pipeline and turned it off in the configs. figured it would be best to keep churn as small as possible.
 - monkey patched method in `mongoid` gem (more on that below)

the one thing that caused issues was the `mongoid` gem. in non-dev environments it tries to preload all the models. it uses `require_dependency` from `ActiveSupport` but it passes incorrect filenames to the method. this caused local rails console and server to run properly but tests would fail. bumping the `mongoid` gem to the appropriate version was going to cause a lot more churn so I decided to override the method in an initializer.
we should keep it around until we upgrade mongo and everything related. it'll require a decent amount of code change (mostly with associations).


zoso10:
branch is failing on CI because #198 is not merged yet. specs will be flakey

igrigorik:
\o/ ... thanks for your work on this!

- Followed up on #198, let's land that first.
- Re, mongo: sgtm.

igrigorik:
Any idea why 2.15 is failing? Odd! Build: https://travis-ci.org/igrigorik/vimgolf/builds/221916683

zoso10:
@igrigorik not sure :/ ran it with the same seed locally and only got it to fail once out of ~20 runs. re-built in CI and it succeeded. the failure that was happening `mongodb ns doesn't exist` means it couldn't find the database or a database wasn't set before querying. im wondering if initial connections fail or something. will keep an eye out for it in the future.

igrigorik:
@zoso10 sg, thanks! Merging.. \o/

PR #200, opened 2017-04-19
Bump rails to version 3.1.12

zoso10:
quick bump of the patch level here. there were A LOT of security fixes from rails version 3.1.0 to 3.1.12. also changed a few gem sources to use `https` over `http` or the insecure git protocol.
hopefully this will make codeclimate a bit happier :)


zoso10:
ugh so it looks like there are still vulnerabilities in version 3.1.12. we fixed 58 issues but introduced 20 more lol. regardless, we're a little better off and we can keep on chugging through upgrading 💪

PR #202, opened 2017-05-05
Bump mongo and related gems to fix deprecation warnings

zoso10:
bump mongo, bson, and bson_ext from version 1.1.5 to version 1.3.1, and also bump mongoid from version 2.0.0.beta.20 to version 2.2.1. this fixes deprecations in `config.generators` and using `class_inheritable_attribute` which will allow us to bump rails versions again. this also allows us to remove the temporary monkey patch to properly load files. I ran this against multiple versions of the mongo database and it all worked correctly still 👍 

we had to remove the association between `User` and `Entry` because `Entry` is nested inside `Challenge` and nested models can't be referenced at the top level like that. i'm assuming the older version of the gem we were using allowed this because it didn't validate associations at loadtime, not really sure.

also added `before_destroy` callback to destroy associated entries. we can't have an association with a `dependent: :destroy` because `Entry` is nested inside `Challenge`. we _probably_ don't need this since there's no place where we `destroy` users, but i didn't want to remove any functionality :)

also removed a spec that that reflected on the association between user and entry, it wasn't doing a whole lot anyway.

I grepped to make sure there weren't any areas where we referenced entries directly from a user ie `user.entries` and couldn't find anything. all access to entries are through `Challenge` and not from the top level 👍


igrigorik:
Did another check for `user.entries` on this end.. and yeah, looks good. Thanks again for the great work!

PR #203, opened 2017-05-05
Bump rails to version 3.2.22.5

zoso10:
this is an easy upgrade. not a whole lot changed from rails 3.1 to 3.2. 3.2 offered some speed improvements in dev mode and routing. there's a few active record changes that won't apply to us since we're using mongoid so I left out those configurations. there's a few deprecations that popped up from this bump so it looks like we'll need to bump the mongo family of gems as a next step.

[here's](http://weblog.rubyonrails.org/2012/1/20/rails-3-2-0-faster-dev-mode-routing-explain-queries-tagged-logger-store/) the official release announcement


igrigorik:
Great stuff, thank you! This brings us to Jan 2012 release -- woo! ;-)

PR #204, opened 2017-05-12
Update mongoid query syntax in prep for major version upgrades

zoso10:
replacing old `find(conditions: {})` syntax with `where().first` syntax. this low of a version of `mongoid` doesn't support the more appropriate `find_by` syntax yet, unfortunately. im doing this now in prep for rails 4.0. also, bump `mongoid` and friends gems to 2.6.0 and 1.7.1 respectively. read through the CHANGELOGs and there's nothing scary in there. did a bunch of manual testing alongside the test suite and it's solid 👍 

doing this in prep for major version upgrades of the `mongoid` and `mongo` gems. the mongo community diverged with the release of mongoid 3.0. that community used their own mongodb driver `moped` which had issues for a while. it wasnt until the release of mongoid 5.0 that mongodb community took over the `mongoid` gem and made `mongo` (what we're currently using) the de facto mongodb driver. little bit of history of the whole thing is written about [here](https://www.mongodb.com/blog/post/announcing-ruby-driver-20-rewrite).

all this matters because we need to bump `mongoid` to at least 4.0 to allow for rails 4.0. i would _really_ like to not mess around with `moped` (since it was always kinda buggy) so im gonna try to bump us all the way to 5.0 where it uses `mongo` gem for mongodb driver again. `mongo` will also be bumped to 2.0 since it was a complete rewrite. this is probably not the proper place for this discussion but i figured id outline my plan in here for right now

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,10 +3,10 @@ ruby '2.1.5'
 
 gem 'rails', '3.2.22.5'
 
-gem 'mongo', '1.3.1'
-gem 'bson', '1.3.1'
-gem 'bson_ext', '1.3.1'
-gem 'mongoid', '2.2.1'
+gem 'mongo', '1.7.1'
+gem 'bson', '1.7.1'
+gem 'bson_ext', '1.7.1'
+gem 'mongoid', '2.6.0'
 
 gem 'json', '1.8.2'
 gem 'memcachier'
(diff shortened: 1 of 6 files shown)


igrigorik:
Thanks for chipping away at this. :)

PR #207, opened 2017-05-22
Bump to rails 4.0.13!

zoso10:
we made it to rails 4! ive been trying to keep the other upgrade PRs smaller and more digestible, but this one grew pretty large. now that i have an idea of all the churn, i could break this PR down if we want.

as outlined in #204 we needed to bump `mongoid` to v5.0 to stay with the mongodb-ruby-driver. `mongoid` went from pessimistic versioning on `activemodel` 3.x to 4.x so rails had to be upgraded at the same time.

ill break down the changes as a result of the mongo and rails version changes separately:

#### mongo changes:
 - `referenced_in`, `references_many`, etc is no longer supported and has now changed to active record's syntax of `belongs_to`, `has_many`, etc.
 - `mapreduce` has been renamed to more idiomatic ruby: `map_reduce`
    - `map_reduce` cannot be called on the meta `collection` object but must be called on the object instance itself
    - `map_reduce` no longer takes an `out` param, but is chained on as a method call. _note_ i used the `replace` strategy because i thought it most closely resembled what the previous behaviour was.
 - `Binary` needs to be prepended with `BSON` namespace
 - there's a completely different config file format, details [here](https://docs.mongodb.com/ruby-driver/master/tutorials/5.2.0/mongoid-installation/#configuration)
 - remove `bson_ext` gem since it's now part of the `bson` gem

#### rails changes:
 - rails 4 introduced strong parameters (syntax: `params.required(:key).value(:foo, :bar)`)
    - we need to use that any time we do mass assignment for a model
    - also added this when we create users from omniauth data in the headers. first we have to wrap it in `ActionController::Parameters` object
 - replaced requiring `active_resource` with `active_model`
 - changed `Bundler.require` to not take options for asset compilation
 - removed javascript expansions configuration (support was completely dropped in 4.0.x)
 - added `via: [:get, :post]` to routes defined with `match`
     - this is not the best solution, but we can go back through and properly define the routes later
 - removed `test_unit` gem so routing specs would pass (this was a fun one to figure out 😄)
 - include each css stylesheet individually in layout
    - i don't think `stylesheet_link_tag :all` works anymore
 - a few other things snuck in from general debugging, they're super low risk so i left them in anyway
    - added `pry_stack-explorer`
    - bumped `rspec-rails` version

~~**Note**: CI seems to be failing from some mongo issue. im guessing it has to do with versions. all the specs pass locally with mongodb version 2.6.x. i also tested locally and everything works/looks great. i'll keep digging but figured i could open up the PR to comments~~
_Update_: looks like setting the version of mongo manually in the `.travis.yml` file worked. i set it to 3.0 which i don't think is what we're running in prodution, correct? (i remember somewhere you said 2.6.5?). i'll keep poking around to see if we can get it set up with 2.6.5 in CI
_Update2_: looks like it was just as easy to get mongodb 2.6 in CI so we're all good 👍


igrigorik:
> Update: looks like setting the version of mongo manually in the .travis.yml file worked. i set it to 3.0 which i don't think is what we're running in prodution, correct? (i remember somewhere you said 2.6.5?). i'll keep poking around to see if we can get it set up with 2.6.5 in CI
> Update2: looks like it was just as easy to get mongodb 2.6 in CI so we're all good 👍

Yep, 2.6.5, but we can update that too.. should, in fact. The only gotcha is I'm not sure if it requires that we update any of the data models.

Gemfile
```diff
@@ -1,12 +1,11 @@
 source 'https://rubygems.org'
 ruby '2.1.5'
 
-gem 'rails', '3.2.22.5'
+gem 'rails', '4.0.13'
```
igrigorik: \o/

app/controllers/sessions_controller.rb
```diff
@@ -19,4 +19,12 @@ def destroy
     session[:user] = nil
     redirect_to root_url, :notice => "Signed out"
   end
+
+  private
+
+  def user_params
+    ActionController::Parameters.new(
+      request.env['omniauth.auth']
+    ).required("info").permit!
```
igrigorik: nitpick.. "" --> ''

app/views/layouts/application.html.erb
```diff
@@ -8,6 +8,16 @@
   <title>VimGolf - real Vim ninjas count every keystroke!</title>
 
   <%= stylesheet_link_tag :all %>
+  <%= stylesheet_link_tag "960" %>
```
igrigorik: what's the point of :all?

config/routes.rb
```diff
@@ -1,22 +1,22 @@
 Vimgolf::Application.routes.draw do
 
-  match "/auth/twitter/callback" => "sessions#create"
-  match "/signout" => "sessions#destroy", :as => :signout
+  match "/auth/twitter/callback" => "sessions#create", via: [:get, :post]
```
igrigorik: Hmm, the fact that we have to repeat this on each route is unfortunate.. =/

app/views/layouts/application.html.erb
zoso10: it used to include all css files under `pubic/stylesheets`. i don't think it works anymore, but couldn't find anywhere in any CHANGELOG that indicated it's no longer a feature. what led me to believe that initially is that there's an ampty `all.css` file so I think after it stopped working rails started trying to serve that file, realized it wasn't there and created it.

regardless, we can probably remove that line for now. my plan is now that we're on rails 4.x we can get asset pipeline up and running, which will do the requiring for us (and also speed up serving assets!)

config/routes.rb
zoso10: yeah sorry, did that initially to make sure we were letting through all the requests we needed for specs. now that everything is passing, i can go through and update the route syntax 👍

app/views/layouts/application.html.erb
igrigorik: sg

igrigorik:
One step closer.. thanks again for your help!

PR #208, opened 2017-05-29
Fix deprecations as a result of upgrading to rails 4

zoso10:
preparing for rails 4.1.x.

the dual secret key (`secret_token` and `secret_key_base`) is so that we can upgrade people's cookies without completely blowing out anything stored there, more details [here](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#action-pack)

we were including `VimGolf` (the gem module) at the top-level in the keylog spec which was causing `VimGolf::Challenge` from the gem to replace the `Challenge` rails model constant. this was causing failures in any rails spec that used the `Challenge` model. we had a few options for resolving this. i decided we didn't need to include the module at all (we could have done it inside the describe block), but instead fully qualify all classes from the gem with the proper namespace. this all manifested when we specifically set `eager_load` configs for all the environments.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -6,9 +6,6 @@ Vimgolf::Application.configure do
   # since you don't have to restart the web server when you make code changes.
   config.cache_classes = false
 
-  # Log error messages when you accidentally call methods on nil.
-  config.whiny_nils = true
-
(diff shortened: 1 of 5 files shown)


igrigorik:
👍

PR #209, opened 2017-05-29
Bump to rails version 4.1.16

zoso10:
this is a pretty unexciting rails version bump. the features added to rails in this version were pretty small and not much was deprecated

_Note_: I can change the base branch once #208 is merged so it'll merge into master. i did it this way so the diff was relevant


igrigorik:
Woot. Great work, as usual.

PR #211, opened 2017-06-01
Bump to rails version 4.2.8

zoso10:
again, another mostly unexciting rails version upgrade. the validation messages changed a little so i had to update a spec


Gemfile
```diff
@@ -40,6 +40,7 @@ group :test, :development do
   gem 'simplecov', :require => false
   gem "codeclimate-test-reporter", "~> 1.0.0"
   gem "capybara"
+  gem "launchy"
```
igrigorik: This is new requirement for 4.2.8?

zoso10: i added that so we could use `save_and_open_page` from capybara for debugging. it's not mission critical so I can remove that if you want :)

igrigorik: Gotcha, ok. Nah, that's fine.

PR #212, opened 2017-06-03
Set up asset pipeline

zoso10:
I was starting to get nervous serving all assets statically and not using rails' asset pipeline during the upgrades. we needed certain gems and configurations to serve static files. further, i have no experience with rails 5 and serving static assets so I figured now would be a good time to set up the asset pipeline. i'm totally open to suggestion if introducing it now is a good idea or not.

it mostly involved moving files and getting a few gem versions correct. the majority of the churn is because i precompiled locally. checking heroku [docs](https://devcenter.heroku.com/articles/rails-asset-pipeline#the-rails-4-asset-pipeline) it looks like heroku will use `public/assets` if it is checked into git otherwise it will do asset compilation when it makes the slug. i made precompilation a separate commit so if we decide against doing it locally we can drop the commit and be fine 👍

i also renamed all css files to `.scss.erb` so we can use sass and have embedded ruby. most of the javascript isn't being used. we use jQuery in one spot in the templates. i removed `video.js` and `rails.js` because they weren't being used. there's a few other misc things that i'd be happy to answer questions about.


config/application.rb
```diff
@@ -57,7 +57,7 @@ class Application < Rails::Application
                     }
 
     # Disable the asset pipeline for now
```
igrigorik: Should probably nuke the comment? :-)

config/environments/production.rb
```diff
@@ -29,7 +29,7 @@
 
   # Disable Rails's static asset server
   # In production, Apache or nginx will already do this
-  config.serve_static_assets = false
+  config.serve_static_files = true
```
igrigorik: Ditto for comment here.

igrigorik:
> i made precompilation a separate commit so if we decide against doing it locally we can drop the commit and be fine 👍

My vote is to drop it and let heroku handle it for us -- minimizes churn on our end.

> most of the javascript isn't being used. we use jQuery in one spot in the templates. i removed video.js and rails.js because they weren't being used. there's a few other misc things that i'd be happy to answer questions about.

Nice. It'd be great to get rid of the jQuery/jquery-UI dependency entirely.. but that's a separate effort.

config/application.rb
zoso10: ah good call!

PR #213, opened 2017-06-06
Bump ruby version to 2.4.1

zoso10:
rails 5 doesn't support any ruby version less than 2.2, so we'll need to bump to a newer version before we can move forward with that. i have it rebased against #212. so i'll keep this blocked by that for now and rebase once it it's merged 👍


igrigorik:
Any reason why we shouldn't move to latest stable in one jump?

zoso10:
was minimizing changes to reduce likelihood of errors. i'm being overly cautious, will bump to latest :)

zoso10:
had to bump `omniauth-twitter` a minor version before bumping `json` gem to 2.0.x to allow for ruby 2.4.x. the [diff](https://github.com/arunagw/omniauth-twitter/compare/v1.2.1...v1.3.0) for omniauth twitter was super unexciting so we should be fine there 👍

igrigorik:
Awesome work, as usual! :-)

PR #214, opened 2017-06-07
Bump rails version to 5.0.3!

zoso10:
in the home stretch now. this upgrade was fairly straightforward as well. a little bit of work getting versions to play nicely, but nothing too crazy. small change in unregistering xml mime type and a few changes to specs because of validations.


igrigorik:
I assume we should land https://github.com/igrigorik/vimgolf/pull/213 first? See my question there.

Gemfile
```diff
@@ -1,11 +1,11 @@
 source 'https://rubygems.org'
 ruby '2.2.7'
 
-gem 'rails', '4.2.8'
+gem 'rails', '5.0.3'
```
igrigorik: \o/

igrigorik:
Woot! We're modern! :-)

PR #215, opened 2017-06-07
Bump rails version to 5.0.3!

zoso10:
sorry, again i didn't have this rebased against master because i was branching off branches, oops! this is basically #214 but going into master this time :)


igrigorik:
Doh, should have caught that..

PR #217, opened 2017-06-10
Verify and prep upgraded version for prod

igrigorik:
Testing new version locally before we deploy it to prod. Ran into a few issues that require small tweaks and updates..


igrigorik:
@zoso10 it looks like the new Auth logic is erroring out, any tips? Feel free to push into this branch.

image

zoso10:
@igrigorik sorry about that, I was being lazy with the permittable params and allowed everything. this was causing extra params like email to try and be set on the user and error. i created a controller spec that exercised the bug and used that to drive the fix in e6709ab

igrigorik:
@zoso10 awesome, thank you! Looks to be working now.

Doing some smoke local tests, ran into another issue:
- On http://localhost:3000/challenges/59384eb3652ee111a0000001
- Submitted a challenge via "GOLFHOST="http://localhost:3000/" vimgolf put 59384eb3652ee111a0000001" - copy paste'd the solution
  - Submit errored out with a 500

```
I, [2017-06-11T09:40:42.924374 #97959]  INFO -- : Completed 500 Internal Server Error in 612ms



ArgumentError (invalid byte sequence in UTF-8):

app/controllers/entry_controller.rb:36:in `block in create'
app/controllers/entry_controller.rb:35:in `create'
  Rendering /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/diagnostics.html.erb within rescues/layout
source=rack-timeout id=1f73ec51113ea44ca8860cc0e4fa41bf timeout=20000ms service=1005ms state=active
  Rendering /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/_source.html.erb
  Rendered /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/_source.html.erb (8.9ms)
  Rendering /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb
  Rendered /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/_trace.html.erb (2.6ms)
  Rendering /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/_request_and_response.html.erb
  Rendered /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/_request_and_response.html.erb (1.3ms)
  Rendered /Users/igrigorik/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.3/lib/action_dispatch/middleware/templates/rescues/diagnostics.html.erb within rescues/layout (75.6ms)
source=rack-timeout id=1f73ec51113ea44ca8860cc0e4fa41bf timeout=20000ms service=1065ms state=completed
```

Interestingly enough, it did the save entry.. When I try to delete it, it also fails with a 500.

image

```
> ruby -v
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
```

Same problem on other challenges. Looks like we have an encoding issue somewhere?

_p.s. I'm testing against production data with `DATABASE_HOST=... rails server -e development"_

zoso10:
@igrigorik ah looks like the `script` column of `Entry` models is causing this error when the validation runs :/ what's weird is that it doesn't fail the validation so it'll persist the record. i'll keep digging and get a fix up

igrigorik:
@zoso10 any luck on your end?

zoso10:
@igrigorik i've been digging, but i'm kinda stumped. it's happening when the challenge saves to the database and is re-writing all the embedded entries. some existing entries cause the error to happen inside validations. ive dug inside that method but it's still not obvious what's causing it.
ive also tried changing `mongoid` versions still with no luck.
we _could_ remove the validation on the `Entry` model since we're already doing a minimum length check (to check for cheating) in the controller. we could add a maximum length check. it's not rails-y but it would prevent these issues.
i'll continue looking either way

igrigorik:
@zoso10 ... it's rewriting all the embedded entries? Really? Wow.. well then, one more thing to fix. :|

Is it worth trying an upgrade to more recent bson, mongoid versions? E.g. https://jira.mongodb.org/browse/RUBY-950 - we're running newer build, but you never know?

If that's a no go, we can remove the validation for now.

zoso10:
sorry! i mispoke. it's *not* updating every embedded entry. the update of the challenge works fine :)

what _is_ happening is that it's querying users after the update of the challenge. one of the users has non utf-8 characters in their name and it's causing the argument error. that explains why the entry saves correctly but the response is still 500, however, it _doesnt_ explain why removing the validation on `Entry` allows it to work :/

igrigorik:
Ah, interesting. Hmm, well.. one solution here is to sanitize the user table and add some user validations and/or tests? Tracking it back down to what caused it may be impossible.. but that's OK, as long as we take some precautions to prevent same sort of errors in the future.

h43z:
Is there still an issue before it can be deployed?

igrigorik:
@h43z yes, I'm seeing crashes due to data issue I described above. We need to figure out a fix for that, and we can then roll it out — which I'd really, really like to see.

h43z:
So as I understand it correctly there are nicknames which are non utf-8? How can one help without the actual data?

h43z:
@igrigorik 
I found new bug that seems to be related. Submitting a solution that has the "backspace" <BS> in it leads to new UTF-8 issues.
My vimgolf solution for a simple test challenge I created was eg `"\x80kbcwworld\e:x\r"` (`\x80kb` being the backspace)
`
MONGODB | localhost:27017 | vimgolf.update | STARTED | {"update"=>"challenges", "updates"=>[{"q"=>{"_id"=>BSON::ObjectId('59dbb018ee9b37490e23f9c6')}, "u"=>{"$push"=>{"entries"=>{"_id"=>BSON::ObjectId('59dc3b33ee9b371a81614563'), "script"=>"\x80kbcwworld\e:x\r", "score"=>12, "created_at"=>2017-10-10 03:14...
MONGODB | localhost:27017 | vimgolf.update | FAILED | String kb is not valid UTF-8. | 0.00017811000000000001s
I, [2017-10-10T05:14:59.403410 #6785]  INFO -- : Completed 500 Internal Server Error in 34ms


  
 is not valid UTF-8.):kbcwworldx
app/controllers/entry_controller.rb:32:in `create'
`
rubys string append `<<` nor mongodb likes the non utf8 conform /x80kb.
Obviously the still in production version of vimgolf is not affected by this. 

In your case the non utf-8 is already within someone else's script entry.

I'm not sure if I'm any help and you already figured that out but you could check if it is indeed a <BS> (/x80kb) issue. 
I was not able to get a solution with backspace into mongodb that I think would lead to your error. I tried.
`
db.challenges.update({...}, {$set: {"entries.$.script": "cw\x80kb\x80kb\x80kbworld\e:wq\r"}})
`
but that did not trigger the error. And the web view would show it very strangly. So this is obviously not how it is stored in prod db.
`cw<0xc2><20-20> <0xc2><BS><0xc2><BS>worlde:wq<CR>`

igrigorik:
> So as I understand it correctly there are nicknames which are non utf-8? How can one help without the actual data?
> My naive solution would be find out which one it is with some query. Doesn't twitter take care of the names in first place? Was there ever a time one could login to vimgolf without twitter?

Nope, and you would think so.. Apparently, somehow invalid utf-8 still snuck in. 

I think you're right, we probably need to write some queries to scrub through the data and fix the bad utf-8 names.

h43z:
@igrigorik can you check my "backspace" theory from above?
The simplest way would be saving a solution to a completely new challenge or one where no solution uses <BS> which if my theory is right should work fine in both cases.

igrigorik:
@h43z I don't see the same problem on live site, but that could be because of old(er) Rails and or Mongo.

h43z:
@igrigorik so can we say that this (meaning backspace entries) cause the error and NOT some invalid usernames or anything other?
Also could you provide me with a challenge or at least entry that triggers the error. I need a faulty one with the backspace in the `script` field.

`	"entries" : [
		{
			"_id" : ObjectId("59dc364aee9b371a8161455f"),
			"script" : "???????",
			"score" : 12,
			"created_at" : ISODate("2017-10-10T02:54:02.394Z"),
			"user_id" : ObjectId("59dbafe5ee9b37490e23f9c5"),
			"updated_at" : ISODate("2017-10-10T02:54:02.395Z")
		},
`

igrigorik:
@h43z apologies about the delay, been offline for the past while.

> Also could you provide me with a challenge or at least one entry that triggers the error. I need a faulty one with the backspace in the script field (as I said I was not able to create one). With that I could try to investigate further.

http://vimgolf.com/challenges/59384eb3652ee111a0000001 was the one that was causing trouble for me. If you ping me on email I can give you a DB dump to investigate.

h43z:
I send you a mail ;)

h43z:
I still think it has something to do with the script entries that contain backspaces (and maybe other weird characters) and **not** the usernames. 
To test this you can use the mongo script below. It changes all script entries of the challenge 59384eb3652ee111a0000001 to the string 'blub'.
```
db.challenges.find({
  _id: ObjectId("59384eb3652ee111a0000001")
})
.forEach(doc => {
  doc.entries.forEach(entry => {
    entry.script = 'blub'
  })unfotunatly
  db.challenges.save(doc)
})
``` 

After that is done submitting with
````GOLFHOST="http://localhost:3000/" vimgolf put 59384eb3652ee111a0000001```
works again. No utf8 error!

Changing all nicknames and names to a valid utf8 string did not make a difference and again that tells me it has nothing to do with the usernames.

I also tried updating every dependency of the stack to the latest version which unfortunately had no positive effect. (utf8 error still present)

Submitting a backspace on the production vimgolf does not lead to an error but in this PR version it does. 

@igrigorik can you point to the commit hash that is used in the production vimgolf version?

**Comment by igrigorik** [comment]
Thanks for the great detective work.. It does sound like you found the culprit. Do you have any recommendations for what we need to fix this moving forward + existing entries?

> @igrigorik can you point to the commit hash that is used in the production vimgolf version?

An ancient one: 52dbdd4.

**Comment by h43z** [comment]
No recommendations for now. I'm pretty much stuck. But I will give it another try.

**Comment by h43z** [comment]
My bug hunting is just not going anywhere. Someone has to look at it with fresh eyes.
I identified two problems. Both within these lines.

https://github.com/igrigorik/vimgolf/blob/69fc731636b12d6e6b7b0061eeca677cc3a721d8/app/controllers/entry_controller.rb#L23-L42

1. Submitting a UTF-8 conform solution will lead to an error when @challenge.save (Line 36) is called

`ArgumentError (invalid byte sequence in UTF-8):`

The easiest way to "fix" this is removing the validation in the entry model
https://github.com/igrigorik/vimgolf/blob/69fc731636b12d6e6b7b0061eeca677cc3a721d8/app/models/entry.rb#L12 

Or another way would be to change the encoding of all non utf-8 entry script's to "ASCII-8BIT".
(You can do this to all or just to the broken ones)
```ruby
        @challenge.entries.each { |e|
          begin
            e.script.split
          rescue
            e.script.force_encoding("ASCII-8BIT")
          end
        }

        @challenge.entries << @entry
```
The .split method does not work on strings like `\x80kbTEST` (\x80kb is a backspace)
```
2.4.1 :004 > "TEST".split
 => ["TEST"] 
2.4.1 :005 > "\x80kbTEST".split
ArgumentError: invalid byte sequence in UTF-8
	from (irb):5:in `split'
	from (irb):5
	from /home/rich/.rvm/rubies/ruby-2.4.1/bin/irb:11:in `<main>'
2.4.1 :006 > "\x80kbTEST".force_encoding("ASCII-8Bit").split
 => ["\x80kbTEST"] 
2.4.1 :007 > 
```

Now deleting challenges work and submitting new solutions with utf-8 valid strings, too.

2. Submitting Non UTF-8 solutions
<Right>, <Left>, <PageDown>, <BS> and more are all keys that can occur in a vimgolf solution and are not valid UTF-8.

Submitting one will lead to an error on line 31 when `@challenge.entries << @entry` is executed.
`ArgumentError (String kbqad2wwplxF,r.0jq3@aZZ is not valid UTF-8.)`
Doing the trick from above by just `force_encoding("ASCII-8Bit")` to the `params[:entry]` won't work.
What **does** work is to uncomment or remove the 
https://github.com/igrigorik/vimgolf/blob/69fc731636b12d6e6b7b0061eeca677cc3a721d8/app/controllers/entry_controller.rb#L29

After that is done `@challenge.entries << @entry` does not throw the ArgumentError anymore but of course it's not a valid entry without a user and therefore the `@challenge.save` will throw a Mongoid Validation error.


I'm not sure what to make of this all. What does `@entry.user = @user` do to the entry thing that `@challenge.entries << @entry` won't work anymore. Unfortunately I have no skill with ruby this is the first time i'm looking at ruby code for more than 10 seconds.

igrigorik:
We definitely want to preserve the original byte information, even if for no other reason then it affords us ability to easily reprocess what we present if/when we make updates to the client. 

>  validates_length_of :script, minimum: 1, maximum: MAX_FILESIZE 

A little confused by this one, why was this one throwing an error for you? Were you testing with a single backspace key, or some such?

--- 

Stepping back, while it's not yet clear what the solution is here, I do think you've identified an important set of test cases we need to address. Would you be up for wiring up these as spec tests? If we have that we can more rapidly iterate through the solution space.

h43z:
@igrigorik i was mostly testing by pressing backspace and then pasting an already working challenge solution.
This always led to `ArgumentError (invalid byte sequence in UTF-8):` 
Does it not behave the same for you?

igrigorik:
I didn't run into that on older builds. Hence my point about getting these codified as tests, instead of manual scenarios — if we can replicate this behavior consistently, we can figure out how to fix it.

h43z:
Don't count on me writing the tests. I will give it a try but this is the first time I'm looking into ruby AND rails.

igrigorik:
TODO: upgrade omniauth dependency (1.3.1 is not live yet).

> The omniauth dependency defined in Gemfile.lock has a known moderate severity security vulnerability in version range < 1.3.2 and should be updated.

igrigorik:
Merging current patches to unblock debugging remaining issues.

PR #221, opened 2017-07-11
Favicon

iovis9:


tpenguinltg:
> What's the benefit of all the various sizes?
Smaller downloads. You want your favicon to be as small as possible.

At minimum, we just need a favicon.ico or a single PNG and specify that in one `link`  tag.

iovis9:
Just following best practices (lots of different platform expect lots of different kinds of favicon). If you want I can just throw the single icon and be done with it.

iovis9:
http://realfavicongenerator.net/faq

igrigorik:
Honestly, the number of files here seems like overkill.. Let's simplify it.

iovis9:
Sure. Any suggestions?

igrigorik:
To keep thing simple, target 1x case with 48x48 icon? I think that's probably sufficient for now.

iovis9:
The 48x48 is just for Android. I'll just do the 16x16 and 32x32.
Can you explain why you feel this is overkill? This is honestly standard favicon support.

PR #232, opened 2018-04-26
Paginate home page

Hettomei:
I updated db/seeds to be able to add lot of entries

So to test with lots of challenge + entries, and to match the 'production' you can do : 

```
bundle exec rake db:setup challenges=400 users=22000 entries=400
bundle exec rake db:setup challenges=1 users=0 entries=20000
```


Hettomei:
Please let me know if you plan to push it to production.

h43z:
@Hettomei I think https://github.com/igrigorik/vimgolf/pull/217 is still blocking.

Hettomei:
@igrigorik 

I have other commits that paginate the leaderboard (right part on a challenge)

What did you prefer?

- I let this PR as it is (easier for code review)
or
- I push here the others commits

Gemfile
```diff
@@ -3,9 +3,9 @@ ruby '2.4.1'
 
 gem 'rails', '5.0.3'
 
-gem 'mongo', '2.3.1'
-gem 'bson', '4.1.1'
-gem 'mongoid', '6.0.0'
+gem 'mongo'
+gem 'bson'
+gem 'mongoid'
```
igrigorik: We have the above frozen to ensure compat with the production version of MongoDB. I assume all of your tests pass with most recent gems? Which version of MongoDB are you testing against?

app/repositories/repository_challenge.rb
```diff
@@ -0,0 +1,69 @@
+module RepositoryChallenge
+
+  def self.collection_aggregate(*args)
+    Challenge.collection.aggregate(args.flatten)
+  end
+
+  def self.count_entries_query
+    collection_aggregate({
```
igrigorik: Could you please add a short description / explainer of what this query is doing, plus docs references to mongo syntax?

app/repositories/repository_challenge.rb
```diff
@@ -0,0 +1,69 @@
...
+          "description" => 1,
+          "created_at" => 1,
+          "count_entries" => {
+            "$size" =>  { "$ifNull": [ "$entries", [] ] }
+          }
+        },
+      },
+      {
+        # create a new field score which is :
+        # (number of entries) / (Time.now - create_at)
```
igrigorik: s/create_at/created_at

app/repositories/repository_challenge.rb
```diff
@@ -0,0 +1,69 @@
...
+    })
+  end
+
+  def self.count_entries
+    result = count_entries_query.first
+
+    result ? result['count_entries'] : 0
+  end
+
+  def self.score_query
```
igrigorik: Nifty query! Ditto, could you provide a short docs explainer of what this does? For future reference and folks that are not familiar with Mongo..

app/views/main/index.erb
```diff
@@ -8,12 +8,14 @@
 <% @challenges.each do |c| %>
     <div>
         <h5 class="challenge">
-            <%= link_to c.title, challenge_path(c) %> - <%= @entry_counts[c.id] || 0 %> entries
+          <%= link_to c['title'], challenge_path(c['_id']) %> - <%= c['count_entries'] %> entries
         </h5>
-        <p><%= c.description %></p>
+        <p><%= c['description'] %></p>
```
igrigorik: Curious, any particular reason why you switched to `[]` syntax?

igrigorik:
@Hettomei thanks again for the great work here, excited to land this and get it in production. One wrinkle is, as @h43z already flagged, the outstanding #217 PR — we need to get that resolved as well.

Re, leaderboard: let's do that as a separate PR, you can fork off this one if you want to run the in parallel.

Gemfile
Hettomei: I removed the version because every things is locked inside Gemfile.lock. so every other dev, and your production website will install the Gemfile.lock during the bundle install

app/repositories/repository_challenge.rb
Hettomei: I'll do that

app/repositories/repository_challenge.rb
Hettomei: Yes

app/views/main/index.erb
Hettomei: All the Model.collection.aggregate return a big hash. It is not mapped to our model because we can add any new fields.

Some month ago I was against that but after playing with other langage I can't see the problem.

But if you are against, and prefer the dot syntax for any reason, I can instanciate a 'Struct' (or any custom object) or maybe try with Challenge.

My preference are 
- keep this hash, (and document method as you ask)
- create a custom Struct tailored for this query
- instanciate a challenge, but I need more time to understand how it behaves with new fields

Hettomei: All the Model.collection.aggregate return a big hash. It is not mapped to our model because we can add any new fields.

Some month ago I was against that but after playing with other langage I can't see the problem.

But if you are against, and prefer the dot syntax for any reason, I can instanciate a 'Struct' (or any custom object) or maybe try with Challenge.

My preference are 
- keep this hash, (and document method as you ask)
- create a custom Struct tailored for this query
- instanciate a challenge, but I need more time to understand how it behaves with new fields

Hettomei:
Thanks for the review, I add the modification. Don't hesitate to impose your point of view :)

Gemfile
Hettomei: And about the version of mongodb, i work with 3.2 but your travis test it on 2.* (Can't remember) so you can be sur this code works on your database 's production version (if you have well configuree teavis.yaml)

igrigorik: I suspect we might run into some issues with the new queries when we run them against old Mongo. That said, the fact that the tests are working a-OK on your end with latest version gives me good confidence that we should do the upgrade. It'll be a tricky deploy.. but we need to do it.

app/views/main/index.erb
igrigorik: Hash is fine, thanks for the context.

igrigorik:
Cool, once we have the docs bits landed we can merge this PR. The deploy will need some extra work, but at least that'll unblock you from submitting the other pagination code.

p.s. thanks again for your help and contributions!

app/repositories/repository_challenge.rb
Hettomei: done

app/repositories/repository_challenge.rb
Hettomei: done

igrigorik:
> I added detailed documentation.

Awesome work, thank you!

igrigorik:
@Hettomei thanks again for the great work on this! 

Let's get the other pagination logic landed as well. :)

Gemfile
Hettomei: I was totally wrong about mongo : 
I said  " i work with 3.2"

No. I work with `MongoDB shell version v3.6.4`
And these query doesn't work under 3.4

You where right about "I suspect we might run into some issues with the new queries when we run them against old Mongo"

And I said "your travis test it on 2.*"
Because I read in .travis.yml `- mongodb-2.6-precise`
But if I read raw log at https://api.travis-ci.org/v3/job/379148525/log.txt
we can see : 
```
MongoDB version
MongoDB 3.4.10
```

That s why this commit worked on Travis

igrigorik: Figures. No worries, I had a strong suspicion this will be the case anyway. We need to bite the bullet and do the upgrade, which is well overdue.

Full speed ahead.

PR #234, opened 2018-05-15
Paginate leaderboard (right part in a challenge)

Hettomei:
Also limit useless computation when requesting challenge using JSON. Faster API ?

Not ready to merge, 
It would be great to have a mongodb master to review query in [app/repositories/repository_challenge.rb  # best_score_per_user](https://github.com/Hettomei/vimgolf/blob/a8b3d04509161fd3a78bb81e56fb9c16f35ba2fc/app/repositories/repository_challenge.rb#L149)

### DIFF
--- a/app/controllers/challenges_controller.rb
+++ b/app/controllers/challenges_controller.rb
@@ -1,3 +1,5 @@
+require_relative '../repositories/repository_challenge'
+
 class ChallengesController < ApplicationController
 
   before_action :login, :only => [:create, :new, :destroy]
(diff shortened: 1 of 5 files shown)


igrigorik:
@Hettomei once again, awesome work! I'm happy to give you access to the Heroku account so you can test drive some of this. Ping me at ilya-at-igvita.com?

igrigorik:
@Hettomei I deployed latest Rails code to the site today. Large challenges (10K+ entries) are timing out, and I'm hoping (fingers crossed) that your work here might help. Let's get this landed? :)

Hettomei:
Ho I see ! congratulation !

I do my best to paginate the right part. Then we have to paginate the solution.

I try to push a mergeable solution this night

Hettomei:
It can be merged.
I checked against production data : the leader are always at the same position :) 


:warning: it is only leaderboard, paginating solution will come in another pr.

app/controllers/challenges_controller.rb
```diff
@@ -49,28 +51,28 @@ def create
...
-            'type' => @challenge.output_type
-          },
-          'client' => Vimgolf::VERSION
-      }}
-
-      format.html
+      format.json { render :json => json_show(challenge_id) }
+
+      format.html {
+        # TODO, there is a better way to do this...
```
igrigorik: Re, TODO: what do you have in mind here?

app/controllers/challenges_controller.rb
```diff
@@ -49,28 +51,28 @@ def create
...
   def show
-    @challenge = Challenge.find(params['id']) rescue nil
-    return redirect_to root_path if @challenge.nil?
-
-    # TODO, there is a better way to do this...
-    users = @challenge.entries.map {|e| e.user_id }.uniq
-    @users = User.where(:_id.in => users).inject({}) {|h,u| h[u.id] = u; h}
+    # Limit to id to avoid downloading uneccessary entries
+    challenge_id = params['id']
+    challenge = Challenge.only(:id).find(challenge_id) rescue nil
```
igrigorik: If there is no matching challenge, isn't this a no-op? Why do we need to do this query and then query again on line 64 for full set of entries?

app/controllers/challenges_controller.rb
Hettomei: It is your TODO :) just moved in previous PR

But I keep it, because we can totally use RepositoryChallenge.paginate_leaderboard to load the page in one query and less ruby

app/controllers/challenges_controller.rb
Hettomei: It is to prevent "format.json" to download full challenge (with all entries) and display only 4 params. I explained it in the commit message https://github.com/igrigorik/vimgolf/pull/234/commits/6ecdc7ec43731aa20d8d312a1bca39aa5b55b618

But still we have a problem in format.html
But it will be solved in another PR that will paginate all the solution.

igrigorik: Ah, I see — thank you.

PR #236, opened 2018-05-23
Cleaning up loose ends to launch Rails 5.x

igrigorik:

PR #237, opened 2018-05-23
Fix submit_entry_spec when POST/entry

Hettomei:


igrigorik:
Should be fixed via https://github.com/igrigorik/vimgolf/pull/236/commits/ea74b2f5c1535dc900151708f9aa6ee48dc435e1.

PR #239, opened 2018-05-28
Fix thousand of useless query per challenge

Hettomei:
In development mode and production mode, was unable to load big
challenge because of tons of

D, [2018-05-28T02:35:48.115951 #25065] DEBUG -- : MONGODB |  | vimgolf.find | STARTED | {"find"=>"challenges", "filter"=>{"_id"=>BSON::ObjectId('55b18bbea9c2c30d04000001')}, "$clusterTime"=>{"clusterTime"=>#<BSON::Timestamp:0x0055a654605f90 @seconds=1527467726, @increment=1>, "signature"=>{"hash"=><BSON::Binary:0x47086434266700 type=gene...
D,

### DIFF
--- a/app/models/entry.rb
+++ b/app/models/entry.rb
@@ -31,6 +31,6 @@ class Entry
 
   # Returns true is an entry was created by given user
   def owned_by?(current_user)
-    user == current_user
+    current_user && user_id == current_user.id
   end
 end


Hettomei:
@igrigorik 

With this PR, combined to #234  I was able to display the big challenge "Simple, Practical, and Common - 17890 entries" on my dev machine.

app/models/entry.rb
```diff
@@ -31,6 +31,6 @@ def self.top_by_user
 
   # Returns true is an entry was created by given user
   def owned_by?(current_user)
-    user == current_user
+    current_user && user_id == current_user.id
```
igrigorik: Where is `user_id` coming from?

Hettomei: It comes from https://github.com/igrigorik/vimgolf/blob/master/app/models/entry.rb#L18
```
belongs_to :user
```
belongs to automatically create `relation_id` https://docs.mongodb.com/mongoid/master/tutorials/mongoid-relations/#belongs-to


I didn't precise : We get all these User.find on Challenge#show : one User.find per printed entry

igrigorik: Gotcha, thanks!

PR #241, opened 2018-05-29
Limit js parsing time on challenge page

### LINKED ISSUE #192: Slow javascript
Since at least several months ago, some piece of javascript on vimgolf.com has been running slow on large challenge pages. I think I've found the problem: a [call to `.button()` in application.js](https://github.com/igrigorik/vimgolf/blob/master/public/javascripts/application.js#L3). Removing it on a local copy removes the issue.

[This link](https://bugs.jqueryui.com/ticket/5454) suggests that initializing a button in jQuery involves an expensive DOM search every time it's triggered, so it causes problems on large pages.

Severity: On the largest pages, this problem causes up to a 30 second delay after `$(document).ready` triggers, after which I can't scroll. (That's on my modern-but-bargain-cheap laptop, running Firefox.) I get up to 3 "Stop script" dialogs during that time. For users who can't see all the solutions yet, it won't be quite so bad, but after they climb the leaderboard...

Hettomei:
Try to fix #192 

Pretty big difference on big challenge : can save 20 seconds. 

# Fake data

## before

before patch, with fake data

## after

after patch, with fake data

# Real data using 'simple practical and common' (4075 golfers)'

## before
before patch-real-data

## after
after patch-real-data


Hettomei:
I removed pretty print because It doesn't add much value to the game but slow down page.

I can remove it from commit and it should still works and win some seconds.

igrigorik:
@Hettomei I'm confused, what's wrong with the current CSS?

image

That looks/works as expected to me, no?

> rspec ./spec/features/entry_feature_spec.rb:57 # Entries for Challenges Entry exists on a Challenge #destroy can delete an entry

Have you looked at the failing spec?

---

Re, remove prettyprint: presumably this will be mostly mitigated once we deploy pagination no? For player ergonomics, I *would* like to preserve the pretty output of the vim script. That said, perhaps there are more performant solutions that we can come up with to address that.

igrigorik:
FWIW, if we're after slimming down JS, I think we could remove jQuery and jQuery UI entirely.. I picked it purely for convenience at the time, but it should be relatively simple to rewrite/remove that code with vanilla JS and CSS — in fact, it's not clear to me that we even need the JS bits?

Hettomei:
Ok, I can try to

• remove jquery/jquery ui
• and let pretty print

And I copy paste button 'style' from jquery.ui.css


Envoyé depuis un smartphone.

Le ven. 1 juin 2018 à 08:56, Ilya Grigorik <notifications@github.com> a
écrit :

Hettomei:
For me it's ready to merge : 

- jquery removed
- css added
- Can't see any difference with previous version
- faster to load page
- tests pass

app/views/challenges/show.erb
```diff
@@ -10,8 +10,8 @@
     <h5>End file</h5>
     <pre class="prettyprint"><%= @challenge.output %></pre>
 
-    <h4><a href="#" onclick="$('.diff').toggle(); return false;">View Diff</a></h4>
-    <pre class="prettyprint diff"><%= @challenge.diff %></pre>
+    <h4><a href="#" onclick="toggle('diff'); return false;">View Diff</a></h4>
+    <pre class="prettyprint" id="diff"><%= @challenge.diff %></pre>
```
igrigorik: `diff` should be a class, since we're repeating it on the page.

Hettomei: I grepped the project and can't see another "diff" (class or id)
Can you show me the file ?

igrigorik: `</facepalm>`

Ignore me. I'm reading diff but thinking submission entry. There is only one diff at the top! All good, merging 👍

PR #243, opened 2018-05-29
Fix #240: a participator cannot delete his comment

### LINKED ISSUE #240: Cannot delete it's own entry
I think it is just a problem of function name (I remember seeing it in dev env, but I m on vimgolf.com...)

PR will come.

Hettomei:
there was an error in method name

### DIFF
--- a/spec/features/entry_feature_spec.rb
+++ b/spec/features/entry_feature_spec.rb
@@ -4,8 +4,6 @@ feature "Entries for Challenges" do
   include OmniAuthHelper
 
   before(:each) do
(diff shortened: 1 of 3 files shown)


spec/features/entry_feature_spec.rb
```diff
@@ -64,4 +66,55 @@
       end
     end
   end
+
+  context 'Entry exists on a Challenge, user is a participator' do
```
igrigorik: Nice! Thanks for going the extra mile to add these.

PR #245, opened 2018-05-30
Fix creating entry on big challenge

Hettomei:
Was unable to create entry on big challenge because of tons of request. (one request per entry)

And so avoid timeout error

the '<<' already perform the 'save'.

The challenge.save do all the request.


Hettomei:
The 'travis fail' is a false positive. You can merge.

Issue #162, opened 2016-02-12
Not working with neovim


GCorbel:
Hi,

I'm using neovim and, when I try to launch vimgolf, I have an error option because `--nofork` is not an option.

urticadioica:
I investigated the options, and it looks like using `set guioptions+=f` in the vimrc could be an alternative. Neovim got rid of that too (according to its help files), but it should ignore the unknown flags. It should really ignore a historical command line option as well, instead of refusing to start, but I guess that's not in line with their design philosophy.

I think there are some differences in what solutions work in a few cases with Neovim. Seems like they got rid of ex mode, for example. But Vim itself changes from version to version, so I wouldn't want to drop support if we can help it.

GCorbel:
I have another PC with vim properly installed and it works. I guess I have to install vim and say to vimgolf to work with it.

Thanks!

letientai299:
I get the same issues today, posted here in case other people need it.
Because I don't want to replace neovim with the original vim, my solution is open the file:
```
vim +113 /var/lib/gems/2.3.0/gems/vimgolf-0.4.8/lib/vimgolf/cli.rb
```
and remove `--nofork` option.
After that, vimgolf and neovim work fine.

majamin:
To extend the answer by @letientai299 : use `gem which vimgolf` to find the location of the gem on your system. This should open it where ever it exists:

`vim +112 $(gem which vimgolf | sed 's/vimgolf.rb/vimgolf\/cli.rb/g')`

PR #246, opened 2018-05-30
Support neovim in vimgolf cli

mayanksuman:
The '--nofork' option is invalid for neovim. 
It is also stated in comment on line 106 that '--nofork' option is only required for gvim.
So, it is more logical to have this option for gvim only not for vim or neovim. 
This code is tested with 'vim', 'neovim' and 'gvim' (vim-gtk3) and it is working without any error.
Solves #162

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ matrix:
     - rvm: 2.4.1 # Test web AND cli with specific version
       gemfile: Gemfile
     # Otherwise test only the cli
-    - rvm: 1.9
-      gemfile: lib/vimgolf/Gemfile
     - rvm: 2.3
       gemfile: lib/vimgolf/Gemfile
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,8 +4,6 @@ matrix:
     - rvm: 2.4.1 # Test web AND cli with specific version
       gemfile: Gemfile
     # Otherwise test only the cli
-    - rvm: 1.9
```
igrigorik: Any particular reason why you removed this?

Hettomei: Because ruby 1.9 is no more supported https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

I m pretty sur it is because since your last merge (upgrade gem) the 1.9 test stopped working because of gem incompatible with r 1.9

igrigorik: Fair enough :-)

igrigorik:
@mayanksuman thank you for the contribution!

mayanksuman:
@igrigorik thanks for merging the changes.

PR #248, opened 2018-06-14
Paginate solutions - reduce memory consumption

Hettomei:
You may not love the 'style' since I do not use Rails magic to create/use object. 
But I cannot find ways to 'not load all entries' when using rails models

# What have been done : 

- Move leaderboard in it's own object
- Create ShowChallenge for various information
- create Solutions to query uniq solution
- paginate solutions (but always display ALL solutions)
- fix pagination : 
  - when not connected
  - when connected with 0 contribution
  - when connected with >= 1 contribution
- fix 'position' of a solution (need to query the 'offset')
- Add tests for queries

# Additionnal infos : 

About 'reduce memory consumption' I use : 

```
pp NewRelic::Agent::Samplers::MemorySampler.new.sampler.get_sample
```

At the beginning and at the end of 'ChallengeController#show'
It looks like it seems close to what heroku returns.

with 
- "Simple, Practical, and Common - 18089 entries" 
- as the best player (display all solution)
- after 10 reload page

Before this PR :
beginning of #show 136, end of #show 191.578125
-> I get lot of requestTimeout, and MemorySampler varie from 110 to 210 

With this PR : 
beginning of #show 136, end of #show 136
-> no requestTimeout, and MemorySampler is stable


app/services/solution.rb
```diff
@@ -0,0 +1,38 @@
...
+    @users = users
+    @position = position
+  end
+  attr_reader :solution, :position
+
+  def id
+    solution[:entry_id]
+  end
+
+  def score
```
igrigorik: Minor, but you can refactor (score, script, comments, created_at) as dynamically defined methods. 

```[x,y,z].each { define_method(...)}```

File: app/views/challenges/show.erb [hunk 2]
```diff
@@ -1,90 +1,73 @@
-<% challenge_id(@challenge) %>
+<% challenge_id(@show_challenge) %>
```
igrigorik: Any reason why we can't keep existing @challenge? It's not clear what `show_` adds in terms of readability.

app/services/show_challenge.rb
```diff
@@ -0,0 +1,41 @@
+require_relative '../repositories/repository_challenge'
+require 'forwardable'
+
+# retieve specific data for show page
+# avoid loading all 'entries' to limit memory
+class ShowChallenge
```
igrigorik: Can't say I'm in love with the name, but struggling to come up with better alternatives.. Any others that you were considering?

app/services/solutions.rb
```diff
@@ -0,0 +1,83 @@
+require_relative '../repositories/repository_challenge'
+require_relative './solution'
+
+class Solutions
```
igrigorik: `Solution` vs `Solutions` is a subtle and easy to miss difference. Perhaps `Solution` and `Entries` or `Submissions` instead?

igrigorik:
@Hettomei this is a big one, kudos for taking it on. Left a few nitpick suggestions above. 

What's left, modulo fixing the failing build?

Hettomei:
> What's left

Implement the system that display solutions step by step (based on the score of players)
Cureently I display all

app/services/solution.rb
Hettomei: For me it is not a refactor, it is a code golf . And as somebody told one day 'I do greppable code'. I think it s better to keep 'def something' .

app/services/solutions.rb
Hettomei: Nice.
I ll try Submissions && Solution

app/views/challenges/show.erb
Hettomei: It was to show to other devs that it is not a 'Challenge model' but 'service/show_challenge' to avoid confusion.

But I agree with you 'ShowChallenge' was a bad name, and so this @show_challenge

Hettomei:
> Implement the system that display solutions step by step (based on the score of players)
--> done

> What's left
- 'count remaining solution'
- add offset to display real solution position
- fix tests
- add tests

Hettomei:
Hi
thanks for the reminder.

I have a working solution on another branch (paginate-solution-2), but I don't like it: 
- too much mongodb query 
- hard to maintain for only one page
- far away from rails style/logic
- zero test (I found edge case that I didn't understood at first, so I should add lot of test to show what I found)

I have an idea to simplify the number/difference/maintenability of queries but I need to take time for that. And I will have time at the end of july

igrigorik:
@Hettomei 👍

Hettomei:
updated description : only test left

app/services/solutions.rb
igrigorik: @Hettomei bump on this one. Could we rename this to Submissions?

app/services/solutions.rb
```diff
@@ -0,0 +1,100 @@
...
+    @solutions ||= RepositoryChallenge.solutions(
+      challenge_id: challenge_id,
+      min_score: visible_score,
+      per_page: PER_PAGE,
+      page: page
+    )
+  end
+
+  def visible_score
+    @visible_score ||= if !player
```
igrigorik: Could we unroll this one into multi-line? Easier to grok :)

app/views/shared/_entries.erb
```diff
@@ -0,0 +1,42 @@
+<% @solutions.each do |solution, user, position| %>
```
igrigorik: if we're renaming to submissions, rename the template as well to _submissions.erb?

Hettomei:
About perf :
In prod env, after 'warming up' the page
I will push what rails gives (ms to display a view)
I will push memory used.

I don't know how to do other good benchmark.

About the review : thanks, i'll update asap.

Hettomei:
# same benchmark, but in a table

|                                  | ram                | page view rails | note                 |
|----------------------------------|--------------------|-----------------|----------------------|
| dev: master; worst player        | 250..256           | OK in 3078ms    |                      |
| dev: pull request; worst player  | 134 then grow 0.26 | OK in 1077ms    | :1st_place_medal:    |
|                                  |                    |                 |                      |
| dev: master; best player         | 272..286           | OK in 7287ms    | sometimes crash :-1: |
| dev: pull request; best player   | 134 then grow 0.26 | OK in 1192ms    | :1st_place_medal:    |
|                                  |                    |                 |                      |
| prod: master; worst player       | 220..238           | OK in 3139ms    |                      |
| prod: pull request; worst player | 116                | OK in 921ms     | :1st_place_medal:    |
|                                  |                    |                 |                      |
| prod: master; best player        | 220..238           | OK in 6682ms    |                      |
| prod: pull request; best player  | 120                | OK in 1147ms    | :1st_place_medal:    |

Hettomei:
Interpretation of the benchmark : 
No need to go deeper, it seems that this PR improve vimgolf.

app/views/shared/_entries.erb
Hettomei: done

app/services/solutions.rb
Hettomei: done

Hettomei:
Again, ready to merge from my point of view ;)

igrigorik:
Awesome work sir. Merging.

PR #252, opened 2018-08-07
Profile page: Add list of played challenge

Hettomei:
Long time ago, we displayed user rank per challenge.
It slowed down application because of heavy data.

I didn't find any solution to simply get the rank from mongo
so I display these information to help user to guess it s rank:

- Best score
- Best player score
- Number of attempts

### DIFF
--- a/app/repositories/repository_challenge.rb
+++ b/app/repositories/repository_challenge.rb
@@ -162,7 +162,7 @@ module RepositoryChallenge
           "created_at": { "$first": '$entries.created_at' },
           "script": { "$first": '$entries.script' },
           "comments": { "$first": '$entries.comments' },
-          "min_score": { "$min": '$entries.score'}
+          "min_score": { "$first": '$entries.score'}
         }
       },
       { "$sort": { "min_score": 1, "created_at": 1 } },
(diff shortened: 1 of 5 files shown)


Hettomei:
It gives something like this : 

localhost_8080_udioica

app/repositories/repository_challenge.rb
```diff
@@ -162,7 +162,7 @@ def self.best_score_per_user(challenge_id)
           "created_at": { "$first": '$entries.created_at' },
           "script": { "$first": '$entries.script' },
           "comments": { "$first": '$entries.comments' },
-          "min_score": { "$min": '$entries.score'}
+          "min_score": { "$first": '$entries.score'}
```
igrigorik: Hmm, for my own education.. Can you explain what the diff is between previous $min and $first here?

igrigorik:
Also, should have said: this is neat! I like the detailed stats - great stuff.

app/repositories/repository_challenge.rb
Hettomei: At the end we get the same result. Just a little optimization and consistancy.

In details : 
Before this $group, we $sort by min_score
When calling $min, it needs to reads all min_score to find the 'min'
But because we sorted it, we can just take the $first

igrigorik: Makes sense, ty!

igrigorik:
Nitpick: can you cleanup the codeclimate complaints before we land this?

igrigorik:
We can add annotations to disable code climate for certain methods, or change the default if we know that this is a recurring pattern. That said, arguably that need not block this PR. Let's tackle that separately.

igrigorik:
p.s. nice work.. again! I'll let you deploy it? :-)

Hettomei:
Code has been deployed \o/

PR #263, opened 2019-01-18
Fix derek wyatt vim tutorials link and use SSL where is possible

onlurking:
This PR will fix the tutorials link from Derek Wyatt in about page:
~http://derekwyatt.org/vim/vim-tutorial-videos~ to http://derekwyatt.org/vim/tutorials
and use the https version of some links where SSL certificate is valid


igrigorik:
Thank you!

PR #264, opened 2019-01-19
use neovim if command is present otherwise use vim

onlurking:
this PR will make vimgolf use [neovim](https://neovim.io) as the default editor if is installed in the computer, the self function is [multiplataform](https://stackoverflow.com/questions/2108727/which-in-ruby-checking-if-program-exists-in-path-from-ruby) so it will run on Windows too.

### DIFF
--- a/lib/vimgolf/lib/vimgolf/cli.rb
+++ b/lib/vimgolf/lib/vimgolf/cli.rb
@@ -1,9 +1,20 @@
 module VimGolf
+  def self.which(cmd)
+    exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
+    ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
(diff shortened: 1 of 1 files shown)


filbranden:
If you prefer to use NeoVim for VimGolf, you can always set the environment variable:

```
export GOLFVIM=nvim
```

You can do so in your bash_profile, bashrc, zsh initialization script, etc.

Personally, I don't think using NeoVim by default if it's enabled is the best approach... But I can see ways in which we could make this easier to select (perhaps in the config file, selected through `vimgolf setup`, etc.) or we could ask the first time we launch, etc.

danilobellini:
Many of the current solutions would not work in neovim, which makes me believe it should not be the default. Also, before enabling neovim, I think some changes to the `.vimrc` would be required as part of the "level playing field" requirement, including something like:

```
if has('nvim')
    unmap Y
endif
```

PR #265, opened 2019-02-24
Remove KE_SNIFF handling.

dstein64:
I've recently implemented a vimgolf client in Python. It has a similar interface as the official client, but adds some features inspired by vimgolf-finder.
https://github.com/dstein64/vimgolf

I browsed the official Ruby code to assist me in my Python version. When browsing the code, I noticed the special handling for `KE_SNIFF`. This is no longer required, as vim now has a `KE_SNIFF_UNUSED` entry in keymap.h.

With the special handling, certain codes will be off-by-1. For example, using the current code on master, if I start a vimgolf challenge, and enter `<C-Left><C-Right>ZZ`, the vimgolf CLI will report the following key sequence:
`<C-Right><C-Home>ZZ`.

This PR fixes the issue by removing the special case handling.

I suppose most users of vimgolf are using the 0.4.8 gem version of vimgolf, which does not have this issue (presumably that version was implemented before vim originally removed `KE_SNIFF`.

### DIFF
--- a/lib/vimgolf/lib/vimgolf/keylog.rb
+++ b/lib/vimgolf/lib/vimgolf/keylog.rb
@@ -28,6 +28,10 @@ module VimGolf
         n = c.ord
         if n == 0x80
           b2, b3 = scanner.get_byte, scanner.get_byte
+          if b2 == "\xfd" && b3 >= "\x38" && @time > SNIFF_DATE
+            # Should we account for KE_SNIFF removal?
+            b3 = (b3.ord + 1).chr
+          end
           code = KC_MBYTE[b2+b3]
           yield code if code # ignore "nil" keystrokes (like window focus)
         else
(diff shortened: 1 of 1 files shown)


dstein64:
My original implementation removed all `KE_SNIFF` handling. This would seemingly work fine for the vimgolf client, but since the server also uses this code to process solutions that were already submitted, I've updated the code so that `KE_SNIFF` special handling takes place during or after April 2016 (the existing `SNIFF_DATE`) and during or before July 2017 (the 8.0.0697 patch that added `KE_SNIFF_UNUSED` was on July 7, 2017).

dstein64:
@igrigorik, thanks for reviewing and merging.

I see a user who has encountered `<C-Right>` appearing as `<C-Home>`:
http://www.vimgolf.com/challenges/5c742a5a50bdf70006d43280#5c74db0250bdf7000cd43652

Assuming the server is running code that has the `KE_SNIFF` special handling, I believe that the update in this PR should address the issue that user is encountering.

igrigorik:
@dstein64 thanks for flagging. We need to release both a new version of vimgolf CLI and bump the server version for that to go into effect. I need to run a few tests and will make it live..

PR #270, opened 2019-06-15
admin addition

braxler:


braxler:
:-) Thanks!

braxler:
@igrigorik, How long does it normally take to deploy a change like this? Does it go out with the next update? Or do I need to do something special now that I've been added?

igrigorik:
@braxler sorry about the delay, should be live now!

Issue #271, opened 2019-07-08
See all entries by the same user for a specific challenge


filbranden:
If a user posts for answers with 157, 156, 155 and finally 151 characters, the challenge page will only show the entry with 151 characters and not the other ones.

The other ones are counted in the total number of entries for that challenge and also on the user page you'd see the user submitted 4 answers.

Would it be possible to show all entries from a single user?

Perhaps if not by default, expose a button to display them all?

Example challenge:
http://www.vimgolf.com/challenges/5c1db460c5949f000990a2f1

 3 active golfers, 6 entries 
(4 are mine)

http://www.vimgolf.com/filbranden
For that particular challenge, shows...
Number of attempts: 4

I think my previous attempts (with 157, 156 and 155 characters) are quite interesting, so I wish they were available as well.

Thanks!

TitouanT:
I agree, it could also be in the form of a timeline to see how people got to the shortest solution. Because sometime there is a back and forth between golfers battling for the 1st place :)
And also it would be nice to keep the comments made on previous solution, there are some pretty interesting discussions sometimes.

Hettomei:
@titouanT maybe you already know it : every comment and every solution are stored. We just need to display it the right way.

If somebody can make a mock up of a clean view ?

I don't have any ideas, but I think theire is something to do with another issue that ask to group 'same entries' of different users.

TitouanT:
@Hettomei I did not knew it but I suspected it, this is good news :)

- In the UI, a button that toggles between "best entry by player" (default) and "all entries of all players" could do the trick
- And the timeline thing could be a another button that toggles between "sort by length (smaller first)" (default) and "sort by date (older first)"
- The per user access could be on the user page, with a challenge id in the url as an optional parameter that would add all the entries in chronological order on the right side of the page (same place as leader-board on a challenge page).
- an alternative to the previous point could be a drop down under each "Played Challenges" to show the player submissions (this one might be better).

@filbranden are my propositions in line with your issue ? (I hope they are ^^)

I think that it would not require too much work since it's all just using what is already there to display entries.

For the grouping of identical entries, maybe just add a "first" label to entries never seen before.
Or add a third option to the first button that would be best "first" entry.

example:
```
Let's assume that two solution of the same length are equal (not true in real vimgolf)
player A, B, C, and D did in chronological order :
A1 -> 22 (first)
A2 -> 19 (first)

B1 -> 21 (first)
B2 -> 18 (first)

A3 -> 18

[...]

PR #272, opened 2019-09-03
WIP : Display all challenge solutions per user

Hettomei:
⚠️ work in progress, no need to review ⚠️ 

This is a partial solution to #271


Hettomei:
PR #293 and #294 superseed this

PR #279, opened 2020-01-08
Add Sakigw to the list of moderators

Sakisan:


config/environment.rb
```diff
@@ -17,7 +17,8 @@
   'federicogalassi',
   'timvisher',
   'udioica',
-  'braxler'
+  'braxler',
+  'Sakigw'
```
igrigorik: I don't believe it should be a problem, but for consistency can we lowercase?

Sakisan: If it is not a problem then sure, but I don't know about that. The URL http://www.vimgolf.com/Sakigw shows the user page while http://www.vimgolf.com/sakigw does not.

igrigorik:
Sanity check: http://www.vimgolf.com/sakigw — doesn't turn up an account?

igrigorik:
Insert triple facepalm here.

http://www.vimgolf.com/sakigw - doesn't turn up an account.
http://www.vimgolf.com/Sakigw - does.

Sorry about the noise, can you revert back to cap?

igrigorik:
Should be live! Thanks for your patience.

PR #280, opened 2020-01-14
Lowercase user name

Sakisan:
Thanks for https://github.com/igrigorik/vimgolf/pull/279. However it appears the lowercase username is necessary after all. 

https://github.com/igrigorik/vimgolf/blob/master/app/models/user.rb#L29


igrigorik:
Rolling out now. Hope it works this time! :)

PR #293, opened 2020-08-09
Include position in ranking for each challenge in user page

filbranden:
Additionally to showing the best player score and the best score for the challenge, include the actual position of the best user score in the ranking for that particular challenge.

Since the query including the rankings is potentially too expensive, only perform the full query and include the rankings if `?ranking=1` is included in the URL, that way just browsing user pages normally will not suffer from the performance hit of the expensive query.

Benchmarking:

I created `db:setup` with `challenges=300 users=10000`, then I patched `db/seeds.db` to have most challenges have a random number of entries in the range 100..2,000, but 2% of them in the range 5,000..40,000, to provide a more realist test scenario. (Creating all challenges with 20,000 entries just made all queries too expensive from the start.)

With that setup, I measured the effects of loading a user page, for a user with entries in 36 challenges, which included all 6 challenges with over 5,000 entries. More specifically, the 6 top challenges ranged from 17,829 to 37,251 entries.

I then loaded this page 10 times, with and without the changes in this PR. I took the average and standard deviation of the 10 measurements for each situation.

 - Without this PR:  0.25547s ± 0.00663s
 - With this PR:     2.23918s ± 0.05312s

This is a roughly 9x increase, which is perhaps quite expensive to make it acceptable for the general case.

PR #296, opened 2020-08-11
Display challenges in user page in chronological order

filbranden:
Make it consistent, so that refreshing the page will not show them in a different order. The ordering matches the one for the challenges at the root page.

Add a spec test for `player_best_scores` with two challenges, which is now possible since the resulting ordering is predictable. (The test is flaky without the corresponding change to the Mongo query.)

Benchmarked this change on a page, timing went from (0.25547s ± 0.00663s) before this change to (0.26706s ± 0.01071s) after it, which I would deem acceptable.

I spawn this from PR #293, with only the commits in that PR which should be safe to merge, since they don't introduce a significant performance regression.

### DIFF
--- a/app/repositories/repository_challenge.rb
+++ b/app/repositories/repository_challenge.rb
@@ -337,6 +337,7 @@ module RepositoryChallenge
           "_id": 1,
           "title": 1,
           "description": 1,
+          "created_at": 1,
           "entries": 1,
           "count_entries": {
             "$size":  { "$ifNull": [ "$entries", [] ] }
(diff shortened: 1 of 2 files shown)


filbranden:
Ping @Hettomei and @igrigorik, can you please take a look?

This makes it so that refreshing a user page won't shuffle the order in which the challenges are displayed...

Hettomei:
Nice :)  It works well for me.
@igrigorik if you merge it, I m available to push it to production, let me know.

Hettomei (approved):

Issue #267, opened 2019-04-10
release a new version


xaltsc:
the latest release is 4 years old and so is the gem.

filbranden:
Bump! Can we please have a new gem? I'm getting the `<fd-61>` thing happening often too...

Hettomei:
@filbranden about the fd-61 problem you can try this workaround if you have docker : https://github.com/igrigorik/vimgolf/issues/268#issuecomment-648723503

filbranden:
@Hettomei The `kramos/vimgolf` Docker image is quite inadequate since it ships Vim 7.4.488 which lacks some important features such as the [`g CTRL-A`](https://vimhelp.org/change.txt.html#v_g_CTRL-A) visual mode command. (Which, BTW, ought to be addressed...)

In any case, the "official" instructions to run it (the three liner at the top right corner) start with `gem install vimgolf`. (Even in cases where Docker is available, sometimes running locally is easier and more convenient, especially where Ruby is readily available.)

It looks like the `<fd-61>` issue was fixed by commit 3bf4116f786148bc49f23056a4fc363425d18cbb, back from 2017 (three and a half years ago.)

Looks like the Gem last release was [version 0.4.8](https://rubygems.org/gems/vimgolf/versions/0.4.8), back in 2015 (five and a half years ago.)

There were [15 commits](https://github.com/igrigorik/vimgolf/commits/master/lib/vimgolf) to `lib/vimgolf` since release v0.4.8.

So I think there are quite a few good reasons to release a new Gem...

Are there any drawbacks of doing so?

Otherwise, can we please get a new release? Pretty please? :-)

Thanks!
Filipe

genebustam:
I agree with this.

PR #301, opened 2020-11-18
Split VERSION into CLI_VERSION (for the Gem) and PROTOCOL_VERSION.

filbranden:
This way it is possible to release a new client Gem without having to upgrade the server to match the new client. (Fix for #267)

Also update Gem dependencies of the client to latest version (at least to solve a deprecation warning in older highlinec

Add a new `vimgolf version` subcommand.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: lib/vimgolf
   specs:
-    vimgolf (0.4.8)
+    vimgolf (0.5.0)
       highline (~> 1.7, >= 1.7.10)
       json_pure (~> 2.1, >= 2.1.0)
       thor (~> 0.14, >= 0.14.6)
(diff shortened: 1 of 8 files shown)


filbranden:
@igrigorik Ping?

I have a follow up that upgrades most of the Ruby Gems on the server side, but it conflicts with this one (due to the changes to `Gemfile.lock`), so I'm holding up on sending that one until this one gets reviewed and merged...

Also, it would be really nice to release a new version of the Gem to rubygems.org, I often bump into the problem with the `<fd>` spurious characters when using the official gem...

Thanks!

filbranden:
Happy New Year @igrigorik !

Ping on this PR? I have a follow up to update the Ruby gems on the server side, but I'd prefer to have this one figured out first. (This one is a change with side effects, since the previous model assumed server and client version in sync on this version, this PR introduces a separate protocol version, so we can push fixes to the client that don't change the protocol without having to upgrade everyone...)

Thanks!
Filipe

filbranden:
Dropped in favor of #308.

PR #308, opened 2021-03-12
Bump up client version to 0.4.9, update dependencies

### LINKED ISSUE #267: release a new version

filbranden:
This is an alternative to #301.

Fixes #267. There have been quite a few important commits that affect the client side and haven't been released.

Also update Gem dependencies of the client to latest version (at least to solve a deprecation warning in older highline.

Add a new `vimgolf version` subcommand.

**NOTE:** The current model requires server and client to be in sync, so the Ruby gem should be published just before and around the same time as the server is updated to require version 0.4.9.

Installing this gem and running it against the current server setup fails with:

```
$ vimgolf put 60340000c06738000924d803
Downloading Vimgolf challenge: 60340000c06738000924d803
Client version mismatch. Installed: 0.4.9, Required: 0.4.8.
	 gem install vimgolf
```

I guess this is expected, as having the server require a specific client is a good way to ensure clients are running the latest version. (Even though in this particular case the changes in the client do not introduce incompatibility with the server.)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,9 +2,9 @@ PATH
   remote: lib/vimgolf
   specs:
     vimgolf (0.4.9)
-      highline (~> 1.7, >= 1.7.10)
-      json_pure (~> 2.1, >= 2.1.0)
-      thor (~> 0.14, >= 0.14.6)
+      highline (~> 2.0, >= 2.0.3)
+      json_pure (~> 2.3, >= 2.3.1)
+      thor (~> 1.0, >= 1.0.1)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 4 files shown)


filbranden:
The one failing test here (from `./spec/lib/challenge_spec.rb:22`) is querying the live server, so the version mismatch is causing the problem... Using a local override to have it query localhost seems to work.

Hettomei (approved):

filbranden:
This is now merged, new gem version pushed to RubyGems, server code pushed to Heroku.

Tested that gem 0.4.8 fails to connect, but a simple `gem install vimgolf` fixes it.

I'll follow up next with a PR to update server-side dependencies.

Thanks!

PR #311, opened 2021-03-22
Change docker image to have an up to date vim and gem version

Hettomei:
Maybe I should have asked to https://github.com/kramos/vimgolf to update the container ?

image : https://hub.docker.com/r/hettomei/vimgolf/

source : https://github.com/Hettomei/docker-vimgolf


filbranden:
Sounds good and thanks for working on it!

Perhaps we should actually eventually merge it into this repository and come up with a GitHub action to release it periodically...

Issue #307, opened 2021-01-29
ActionView::Template::Error (Sort exceeded memory limit of 33554432 bytes, but did not opt in to external sorting.


Hettomei:
Using `heroku logs -t | grep -i error`

We can see a lot of 

```
2021-01-29T21:33:28.922622+00:00 app[web.1]: ActionView::Template::Error (Sort exceeded memory limit of 33554432 bytes, but did not opt in to external sorting. (292)):
2021-01-29T21:33:29.011488+00:00 app[web.1]: I, [2021-01-29T21:33:29.011393 #9]  INFO -- : Completed 500 Internal Server Error in 819ms
```

Issue linked to #306

Hettomei:
@filbranden do you think it can be because of https://github.com/igrigorik/vimgolf/commit/0828afaf1b858fc9cfa9e57106abecf1fd172e51 ?

filbranden:
That code path should only trigger if you query a page with `?ranking=1`, so I'd imagine it's not due to that commit and you might be able to double check for that by looking at the access log and seeing whether queries including that parameter have been issued (assuming that's logged somewhere.)

Having said that, the main reason why that commit didn't work is the same as this one, the "Sort exceeded memory limit". While that one will consistently run into this issue for any non-trivial user, I guess as users post more entries in more challenges, the sort limits will probably get reached in those cases too...

I also pushed 8a1163235d4631300ec5fb32acd33c01c616cc56 which should solve this issue, as it addresses the "but did not opt in to external sorting" part, at least that's what Stack Overflow tells me, and as far as I can tell all MongoDB queries go through that code... But for some reason that doesn't really work...

One thing I would like to do is upgrade the Ruby gems to newer versions, I actually have a stack of commits locally to do that, maybe newer versions of the gem will have a fix for the problem? They build upon #301 and that one has been pending review for a while...

I just read last week that @igrigorik got a new job (Congrats!) so I guess he's been pretty busy and will probably be busy ramping up for a while... I'd be happy to try to push these PRs through. BTW, I'd be glad to help as a maintainer if you'd take me (both on GitHub, on Heroku), feel free to contact me by e-mail if you'd consider taking me into the team.

Hettomei:
Thanks for the insight

I extracted 8 min of logs, here are the path and the errors 
[list_url.txt](https://github.com/igrigorik/vimgolf/files/5896749/list_url.txt)

filbranden:
[...]

PR #314, opened 2021-04-05
Try to fix "Sort exceeded memory limit of 33554432 bytes"

Hettomei:
About #307

Maybe I miss something VERY obvious (in the way that it didn t fix anything) but I got good result on my computer with these 2 lines.

before starting server, (and also before deploying on heroku), you need to do a 

```
bundle exec rake db:mongoid:create_indexes
```

Let me know if it works for you


Hettomei:
So yes, it broke something because unit test says : 

```
   expected: 29
            got: 52
```

but maybe it is acceptable on a 'degraded-experience' branches ?

Hettomei:
This leads me to a new idea. I close this Pull request since the unit test tell me there is a problem.

Issue #316, opened 2021-04-07
Twitter images not working


filbranden:
Twitter avatars using `https://avatars.io/twitter/#{user}/small` are no longer working.

@igrigorik switched to `https://twitter-avatar.now.sh/#{user}` in the git deployed to Heroku, but it seems that one stopped working too.

I don't really see any service able to serve a Twitter avatar from a username... I'm guessing we're now supposed to query the Avatar URL from the profile and cache it locally, then serve it? Not sure how that works with profile picture changes though...

We might want to look for an alternative while that's broken...

cc @Hettomei

PR #317, opened 2021-04-07
Switch Avatar to use an Identicon while Twitter profile pics are unavailable

filbranden:
Switch Avatar to use an Identicon while Twitter profile pics are unavailable.

I found this service which provides a consistent identicon based on the passed information. Pass it an MD5 digest of the username as the unique information.

The Identicons are not as good as an actual profile picture... But they're definitely better than serving a broken image (or removing the avatars altogether.)

This is a workaround for #316 until we have a better solution for it.

cc @Hettomei

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -8,7 +8,7 @@ module ApplicationHelper
   end
 
   def twitter_avatar(user)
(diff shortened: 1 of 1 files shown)


Hettomei:
OK for the idea + merge

PR #318, opened 2021-04-07
Switch to using Challenge#top_entries instead of Mongo queries.

### LINKED ISSUE #307: ActionView::Template::Error (Sort exceeded memory limit of 33554432 bytes, but did not opt in to external sorting.

filbranden:
Switch to using Challenge#top_entries instead of Mongo queries.

MongoDB queries were hitting sorting limits and disk use for sorting isn't available in the MongoDB Atlas tier we're using.

So let's switch to just accessing Challenge#top_entries and do the filtering in Ruby code. This is a little bit slower, but it should work just fine for reasonably sized challenges. In my testing, I had trouble with a 37,000 entry challenge (request exceeded 15,000ms) but up to ~30,000 entries it should work fine.

This should help fix #307.

This is now deployed to https://vimgolf-staging.herokuapp.com/ (but for some reason it's timing out there, I need to take a closer look at what's happening there.)

cc @Hettomei

### DIFF
--- a/app/models/challenge.rb
+++ b/app/models/challenge.rb
@@ -23,7 +23,7 @@ class Challenge
   validates_length_of :diff, minimum: 1, maximum: MAX_FILESIZE
 
   def top_entries
-    entries.top_by_user
+    @top_entries ||= entries.top_by_user
   end
 
   def participator?(current_user)
--- a/app/repositories/repository_challenge.rb
+++ b/app/repositories/repository_challenge.rb
@@ -247,12 +247,14 @@ module RepositoryChallenge
   # RepositoryChallenge.best_player_score(challenge_id, user_id).to_a
   # => 123
   def self.best_player_score(challenge_id, player_id)
-    result = collection_aggregate(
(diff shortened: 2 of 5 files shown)


filbranden:
Fixed the issue with staging, it was related to being logged in. I managed to open a challenge with 5,713 entries, but the one with 10,361 entries timed out... Still, I guess that's something...

filbranden:
I found this was not completely satisfactory, so let's go with the approach in #315 instead.

filbranden:
No longer relevant after #322.

PR #322, opened 2021-04-29
Migrate Vimgolf from Mongoid to ActiveRecord!

### LINKED ISSUE #298: Would you consider migrating away from MongoDB?
cc @Hettomei 

Based off of some benchmarks in #293, I started looking for alternative approaches.

One that occurred to me was to use store the rank in MongoDB and use transactions to keep them up to date, but after initial testing I quickly realized that transactions in MongoDB require specific setups of MongoDB, for instance it doesn't seem to work with the local development/testing setup explained in the readme...

So I started thinking... Why keep using MongoDB? The data itself is quite well suited for a relational database and that's something even easier to support in Rails (pretty much natively) than using MongoDB...

I also saw that Postgres and Sqlite both support a [RANK() function](https://www.postgresql.org/docs/current/tutorial-window.html) that we could use for figuring out the rank of each solution, which is also pretty nice...

Other advantages of using Postgres + Sqlite are that it's much easier to run a local development environment (pretty much stock rails does everything) and I imagine it might be simpler/cheaper to host the solution without the addition of MongoDB.

I started working on this, and while I can see that it's a considerable amount of effort, I think it's pretty doable and I'd be willing to put some time into making that work...

But before I take the plunge... Would you consider migrating away from MongoDB to something like Postgres + Sqlite?

### LINKED ISSUE #306: Some challenges failing to load
Some challenges are failing to load. Only pattern I've noticed is that the higher the entries the more likely it is to fail.

One example:
http://www.vimgolf.com/challenges/5192f96ad8df110002000002
"We're sorry, but something went wrong."

filbranden:
Create an initial migration to create VimGolf tables. These will hold the Users, Challenges, Entries and Comments. They mostly map the data stored in MongoDB collections.

The Challenges table also include a column for `legacy_urlkey` that holds the MongoDB id of a specific challenge. Since that MongoDB id is currently exposed as the URL of a challenge (or `vimgolf put` argument), we will preserve it during the migration to keep URLs stable.

Comments will now store a reference to the Users table, rather than only storing the textual nickname of the user who originally posted the challenge (nicknames can actually change.)

Ideally some of the fields (such as Users.key and Users.nickname) should have a unique index, but there are some data issues with repeated entries for the same nickname or key, which we will address in a follow-up after the migration from MongoDB.

Update the code to use ActiveRecord models, rather than Mongoid documents, to implement the entities stored by the application.

For ranking entries (only the top entry by each user is presented), we are using SQL window functions, which allow us to execute the bulk of that filtering on the database side. Method `Challenge#top_entries` implements query to the ranked entries for a challenge in a form that is suitable for our purposes.

Window functions on PostgreSQL:
https://www.postgresql.org/docs/current/tutorial-window.html

Window functions on SQLite3:
https://sqlite.org/windowfunctions.html

(NOTE: Window functions are only available in sqlite3 version 3.25+, so developing and testing the app requires a system with a recent enough version of sqlite3. For instance, Ubuntu 18.04 LTS ships sqlite3 3.22, which doesn't have the necessary support. Ubuntu 20.04 LTS ships an adequate sqlite3 version.)

The MongoDB ids were exposed in the URLs for Challenges, so in order to keep the URLs stable and keep backwards compatibility, we introduce a new `urlkey` to encode the key to use for a Challenge URL.

For challenges migrated from MongoDB, column `legacy_urlkey` in the database will hold that information.

For newly created challenges, we are using a 24-character similarly looking key, using the `9v0` prefix, followed by 9 hex digits encoding the `created_at` timestamp (so that URLs will sort in creation order) followed by 13 hex digits encoding the ActiveRecord sequential id of the Challenge. Prefix `9v0` is used because it's higher than any MongoDB ids (which currently start with `4` and `5`), so it sorts higher, and the `v` makes it look similar while not matching a hex digit. The trailing `0` allows for some extension to change to a different scheme in the future again, while still being able to preserve URLs and keep ordering.

Another difference from the MongoDB implementation is that the home page will now display Challenges in strictly chornological order, newest to oldest. The old code tried to prioritize "popular" challenges, but the way I see it there were some shortcomings with that approach, for instance a newly created Challenge would typically end up drifting to a later page and not be visible enough. (We can address highlighting "Popular" or "Featured" challenges in a follow-up.)

Finally, the MongoDB implementation would simply store the nickname of a user in a Comment field. We're replacing that with a proper foreign key to the User who initially posted the comment, in order to preserve referential integrity.

This setup using ActiveRecord was tested locally, with a full import of production MongoDB data from a couple of weeks ago, and proved to be very responsive.

As an example, rendering the leaderboard for challenge "Simple, Practical, and Common", with 32,662 entries, took around 600ms.

Rendering a user page for that same challenge, displaying all entries by a single user, took under 300ms.

Here are the logs for these two accesses:

```
I, [2021-04-29T11:29:16.460899 #1106821]  INFO -- : Processing by ChallengesController#show as HTML
I, [2021-04-29T11:29:16.461014 #1106821]  INFO -- :   Parameters: {"id"=>"55b18bbea9c2c30d04000001"}
I, [2021-04-29T11:29:17.022771 #1106821]  INFO -- : Completed 200 OK in 562ms (Views: 70.8ms | ActiveRecord: 486.7ms)
127.0.0.1 - - [29/Apr/2021:11:29:17 -0700] "GET /challenges/55b18bbea9c2c30d04000001 HTTP/1.1" 200 - 0.5750

I, [2021-04-29T11:30:59.762671 #1106821]  INFO -- : Processing by ChallengesController#user as HTML
I, [2021-04-29T11:30:59.762731 #1106821]  INFO -- :   Parameters: {"id"=>"55b18bbea9c2c30d04000001", "username"=>"filbranden"}
I, [2021-04-29T11:31:00.033980 #1106821]  INFO -- : Completed 200 OK in 271ms (Views: 29.7ms | ActiveRecord: 239.8ms)
127.0.0.1 - - [29/Apr/2021:11:31:00 -0700] "GET /challenges/55b18bbea9c2c30d04000001/user/filbranden HTTP/1.1" 200 - 0.2776
```

Update tests to use the ActiveRecord models, also Travis-CI config to use Ubuntu 20.04 LTS (required due to the use of sqlite3 window functions.)

Finally, add a rake task `mongodb:import` to migrate data from MongoDB. The task uses direct access to the MongoDB collections and creates instances of ActiveRecord classes to import the data.

This task takes an optional `challenge_ids=...` argument, containing a comma-separated list of MongoDB ids, in order to only migrate those specific challenges. (Note that, in either case, the `mongodb:import` task requires the database to be initially empty.) When importing only a subset of all challenges, the task will also only import users on demand as they're referred to by the challenge, entries or comments, so the list of imported users will also be a partial one. This mode of operation is useful to test real data on the Heroku Postgres free tier, which has a limit for 10,000 rows, so wouldn't fit all entries (even all entries of single very popular challenges), or all active golfers.

The `mongodb:import` task also handles connecting Comments back to Users through a direct relationship, looking up the user by its nickname and storing the ActiveRecord id for that user in the newly imported Comment.

Since nicknames can change (they're based on Twitter handles and users are allowed to change those), some comments refer to nicknames that are no longer valid users. In order to handle those during the migration, we're creating "dummy" new users, using a provider `none` to tag these cases. (We might need a follow-up with some data cleanup here, after the migration out of MongoDB.)

The `mongodb:import` task also collects and reports statistics on the data migration, to help inspection of progress and monitoring the performance of the operation.

Tested migrations locally (to both SQLite3 and PostgreSQL) and on Heroku with Postgres free tier (limited to 10,000 rows):

  1. Migrate 8 selected challenges into local SQLite3 database:

```
I, [2021-04-28T15:48:45.681949 #1084283]  INFO -- : MongoImporter: Migration *COMPLETED*!

--------------------------------------------------
Users..............:    9098
Challenges.........:       8 /       8 (  1%)
Entries............:   65545
Comments...........:     152

Total Objects......:   74803
Elapsed............:   5min 58sec
--------------------------------------------------
```

   2. Migrate all challenges from a copy of the production MongoDB database as of a few weeks back, into a local SQLite3 database:

```
I, [2021-04-28T16:23:46.085219 #1085304]  INFO -- : MongoImporter: Migration *COMPLETED*!

--------------------------------------------------
Users..............:   30687 /   30687 (  1%)
Challenges.........:     513 /     513 (  1%)
Entries............:  372776
Comments...........:    3141

Total Objects......:  407117
Elapsed............:  32min 27sec
--------------------------------------------------
```

   3. Migrate all challenges from a copy of the production MongoDB database as of a few weeks back, into a local Postgres database:

```
I, [2021-04-28T18:05:53.863711 #1094634]  INFO -- : MongoImporter: Migration *COMPLETED*!

--------------------------------------------------
Users..............:   30687 /   30687 (100%)
Challenges.........:     513 /     513 (100%)
Entries............:  372776
Comments...........:    3141

Total Objects......:  407117
Elapsed............:  14min 32sec
--------------------------------------------------
```

  4. Migrate selected challenges, directly from the production MongoDB instance, into a Heroku Postgres free tier database, with the migration running on Heroku:

```
$ heroku run --app vimgolf-staging bundle exec rails mongodb:import challenge_ids=5f1e0217becb80000692b9c4,5ebe8a63d8085e000c2f5bd5,6013804df3308e0009368f1c,51a388ac06f7d20002000006,592e0f555b60ba0009000001
...
I, [2021-04-29T01:26:38.586319 #4]  INFO -- : MongoImporter: Migration *COMPLETED*!

--------------------------------------------------
Users..............:    1062
Challenges.........:       5 /       5 (100%)
Entries............:    4384
Comments...........:      35

Total Objects......:    5486
Elapsed............:   1min 26sec
--------------------------------------------------
```

The expectation is that the real data migration from MongoDB to Heroku Postgres importing all data should take around 15 minutes (similar to local Postgres) and shouldn't take much longer than 30 minutes.  Reporting will allow for monitoring progress of that migration, allowing us to back off if it seems something is going unexpectedly.

Fixes #298. (Also fixes #306 and #307.) Cc @igrigorik and @Hettomei.


filbranden:
Talked to @igrigorik offline about going ahead with this one.

I'll merge it now, push it a little later today and perform the data migration to Postgres.

I'll post back here once it's running in production.

PR #324, opened 2021-05-07
Update README to reflect new development backend on Active Record

filbranden:
I dropped the instructions regarding MongoDB and added the instructions for SQLite3 and Postgres.

Also rephrased or clarified parts of the README that I thought could benefit from it.

I updated `database.yml` to include both SQLite3 and Postgres setups for development + test and use an environment variable (`DATABASE_ADAPTER=pg`) to select using Postgres in development. Also disabled dumping schema to `db/schema.rb` when Postgres is in use, as the formats seem to be somewhat incompatible. These are all reflected in the documentation.

Feel free to take it for a spin and comment or rephrase anything you'd like to have changed.

/cc @Hettomei

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,12 @@ $> vimgolf setup
 $> vimgolf put [challenge ID]
 ```
 
+## Golfing without ruby installation: [use docker](https://hub.docker.com/r/hettomei/vimgolf/)
+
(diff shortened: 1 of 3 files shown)


Hettomei:
Thank you

PR #326, opened 2021-05-14
Include position in ranking for each challenge in user page

filbranden:
This reimplements PR #293 on top of ActiveRecord.

Also re-enable the test cases that were temporarily disabled. Two of those were testing for this feature, the third one tested a related feature (showing number of submissions for the same challenge) but it was using a method removed during the ActiveRecord migration.

Tested:

- Accessing a user page with `?ranking=1` displays the position of the best user submission for each challenge.
- Tests under `bundle exec rspec` passing.
- Ran rspec tests against both SQLite3 and Postgres backends, both passed successfully.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -18,13 +18,24 @@ class User < ActiveRecord::Base
 
   def player_best_scores
     Entry.from(
-      Entry.where(user_id: id).select(
-        '*',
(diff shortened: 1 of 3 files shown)


filbranden:
See example on staging website on Heroku (with a limited database, not all challenges present, snapshot from a few weeks back):
https://vimgolf-staging.herokuapp.com/filbranden?ranking=1

PR #329, opened 2021-05-21
Tidy up, run app:update, upgrade gems.

filbranden:
A few commits tidying up and updating files and gems.

A while ago I updated the gems from Rails 4.x to Rails 5.x, but I missed the step of running `rails app:update` to sync up the config files to the latest version of Rails, so I did that now. This uncovered the issue with a test (validation of belongs_to is now performed by default), so I fixed the test as well.

I also upgraded the rails-rspec gem from 3.x to 4.x, according to their README version 4.x should be used with Rails 5.x or 6.x. Also ran a general gem upgrade of everything.

Fixed a deprecation warning on SQLite3 adapter, related to storing boolean values. Since we don't have boolean values in the database, tweaking the option shouldn't affect us.

So, all in all, a bunch of changes, but mostly no side effects expected.

Deployed the app to https://vimgolf-staging.herokuapp.com/ and confirmed everything is working as expected.


Hettomei:
I just read the code but everything looks like what you described. So, off course, merge when you want.

filbranden:
Thanks for taking a look @Hettomei! Have a great weekend!

PR #333, opened 2021-05-26
Release new client gem 0.5.0

filbranden:
This includes the Net::HTTP changes to use https by default.

Also add explicit requirement on Ruby >= 2.0 for the client gem.

We have one test that hits (actual!) production, so in order to work with it set GOLFHOST to point to the staging environment. I pushed this version to the staging environment already, so the test should work with it.

I already added a changelog entry for this Friday, I'm expecting to push the new version on that date...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: ruby
 os: linux
 dist: focal
+env:
+  - GOLFHOST=https://vimgolf-staging.herokuapp.com
 jobs:
   include:
     - rvm: 2.6.7 # Test web AND cli with specific version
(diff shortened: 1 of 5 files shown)


filbranden:
I'll go ahead, merge and release that.

(Note that the complaint from code climate doesn't make much sense, since all this is doing is updating the version constant, the gem lock file and the changelog...)

I'll release the new gem to rubygems and push the site shortly. (This means everyone will have to update to the latest gem too.)

I'll enable "force https" on Heroku, I'm expecting it will work.

Heads up @Hettomei, you might want to release a new container with the updated client once this is released. Thanks! (Also, looking into automating that step, for the future.)

Hettomei:
new image has been uploaded to hettomei/vimgolf:latest

thanks for the ping

PR #338, opened 2021-06-09
Force HTTPS everywhere.

### LINKED ISSUE #299: Redirect HTTP to HTTPS to increase security
Currently, the website `www.vimgolf.com` can be visited via either HTTP or HTTPS protocols.

Feature request: automatically redirect HTTP to HTTPS with status code 301 at the server level.

filbranden:
Use Rails-provided config.force_ssl to redirect to the HTTPS version whenever the HTTP version is hit.

Furthermore, switch our identicon source to use HTTPS instead, to prevent any warnings about mixed content.

Tested on staging environment: access http://vimgolf-staging.herokuapp.com/ and confirm it redirects to HTTPS. Also checked that the identicons are served using HTTPS and no mixed content warnings are displayed on the browser.

(NOTE: The production environment is slightly different from the staging one, in that for the production HTTPS is done by Cloudflare, redirecting to HTTP on Heroku, while on the staging environment both HTTP and HTTPS are served by Heroku. Even then, the Rails force_ssl with the redirect should still work fine on both. In any case, if something goes awry we can always rollback the version to prior to this commit and come up with a different approach if needed.)

Fixes #299.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -42,7 +42,7 @@ Rails.application.configure do
   config.active_storage.service = :local
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.


filbranden:
@igrigorik I tried to push this to vimgolf Heroku in production, but it produces an infinite redirect loop.

I strongly suspect this is related to the Cloudflare SSL that's hosting the vimgolf.com domain in front of the Heroku app.

I did some research (for example, [this question on stack overflow](https://stackoverflow.com/questions/32956705/force-ssl-redirect-loop-on-rails-4-using-cloudflare-ssl)) and I believe the solution is to change Cloudflare SSL to use mode "Full" instead of "Flexible":

https://developers.cloudflare.com/ssl/origin-configuration/ssl-modes/#full

Since Heroku has HTTPS working for the domain that hits Heroku directly (https://vimgolf.herokuapp.com/), I believe "Full" mode on Cloudflare should keep working right away and would allow Rails to use `force_ssl`.

@igrigorik Would you be able to switch the Cloudflare SSL configuration from "Flexible" to "Full" please? Thanks!

Alternatively, we could point the domains directly at Heroku, since it does have HTTPS support already anyways, not sure if there's that strong of a reason to keep a Cloudflare proxy in front of it.

igrigorik:
Done. Give it a try again?

image

filbranden:
This worked as expected.

Tested on staging environment, also pushed to production and confirmed it works there, behind Cloudflare SSL in "Force" mode.

For the production instance, confirmed that the http://vimgolf.herokuapp.com endpoint which hits Heroku directly (not via Cloudflare SSL) now properly redirects to HTTPS as expected, so I can tell the Rails force_ssl is really working.

filbranden:
BTW, it looks like accessing `https://www.vimgolf.com/` directly on Heroku works fine, with SSL, with a certificate from Let's Encrypt that Heroku is managing on our behalf:

```
% curl \
        --no-progress-meter -v \
        --connect-to 'www.vimgolf.com:443:agile-goat-oe2s1yx2yh0um9bag6ocymeo.herokudns.com:443' \
        https://www.vimgolf.com
* Connecting to hostname: agile-goat-oe2s1yx2yh0um9bag6ocymeo.herokudns.com
* Connecting to port: 443
*   Trying 52.202.168.65:443...
* Connected to agile-goat-oe2s1yx2yh0um9bag6ocymeo.herokudns.com (52.202.168.65) port 443 (#0)
...
* Server certificate:
*  subject: CN=www.vimgolf.com
*  start date: Aug 23 02:59:09 2022 GMT
*  expire date: Nov 21 02:59:08 2022 GMT
*  subjectAltName: host "www.vimgolf.com" matched cert's "www.vimgolf.com"
*  issuer: C=US; O=Let's Encrypt; CN=R3
*  SSL certificate verify ok.
> GET / HTTP/1.1
...
< HTTP/1.1 200 OK
< Server: Cowboy
...
< Via: 1.1 vegur
<
{ [2372 bytes data]
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"
    "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en-us">
...
```

Note that name `agile-goat-oe2s1yx2yh0um9bag6ocymeo.herokudns.com` is the one provided by Heroku for the VIPs serving this virtual host (see the "Domains" section under [Settings](https://dashboard.heroku.com/apps/vimgolf/settings)).

So not sure how much the Cloudflare SSL is still adding to this solution, but since Heroku seems to be providing HTTPS successfully already, perhaps dropping Cloudflare SSL and updating the DNS to point at the Heroku names directly would be a good option. Let me know if you have any interest in considering that.

igrigorik:
I think I'd prefer to keep CF in front, as it offers other options & benefits as a CDN / proxy.

Issue #304, opened 2021-01-10
Neovim 0.5 removes -Z flag


dstein64:
The next Neovim release, version 0.5, [removes](https://github.com/neovim/neovim/issues/11972) the `-Z` flag (restricted mode), which vimgolf [uses](https://github.com/igrigorik/vimgolf/blob/0828afaf1b858fc9cfa9e57106abecf1fd172e51/lib/vimgolf/lib/vimgolf/cli.rb#L112) to disable the use of external shell commands and other functionality.

This will break vimgolf for users that use Neovim (Issue #162, PR #246) with e.g., `GOLFVIM=nvim`.

filbranden:
Bummer!

BBerabi:
is there any way to fix this?

lennartack:
As a temporary solution I'm using a wrapper:
```
#!/bin/bash
exec nvim -n --noplugin -i NONE +0 -u "${@:(-6):1}" -W "${@:(-2):1}" "${@:(-1)}"
```
That said, I do think it's cheating if your solution doesn't work on vim.

PR #346, opened 2022-04-25
Remove -Z when GOLFVIM=nvim

AlexApps99:
See #304

This PR conditionally removes the `-Z` flag when `$GOLFVIM` is set to `nvim`.

This may make it slightly easier to cheat, but it doesn't really change that much considering that you can make a shellscript for `$GOLFVIM` that cuts out the first arg pretty easily. (`exec nvim "${@:2}"`)

I also changed the diff tool to use the provided vim by default, since most people that change `GOLFVIM` would expect vimgolf to use it for diff too.

Making the experience for nvim slightly better.


AlexApps99:
For good anti-cheat/validation, it might make sense to validate the key presses on the server (no idea how that'd work, unlikely that would ever be implemented, just putting it out there)

jenvigo:
I would love to see this PR merged.

PR #353, opened 2022-08-23
Use ruby_identicon to produce an identicon as a data URL

filbranden:
This removes an external dependency that has been broken for a while.

The ruby_identicon gem produces a PNG image in a format that is suitable to use in a `data:` URL, which also reduces the number of HTTP connections needed to properly render a page.

Tweak the identicon size to match our 48x48 space reserved for images.

Tested: Actually pushed to both vimgolf-staging and production already.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem 'sprockets-rails', '~> 2.0'
 gem 'sass-rails', '~> 5.0'
 gem 'uglifier', '>= 1.0.3'
 gem 'kaminari'
+gem 'ruby_identicon'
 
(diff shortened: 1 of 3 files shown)


filbranden:
CI is failing on Ruby 2.0, I imagine pinning the version of one of the test gems in the client library should fix that, I'll take a look at that soon. In any case, I think we don't need to block this change, since the Rails app uses Ruby 2.6 on Heroku anyways (BTW, we should probably upgrade to 2.7 or 3+, I'll look into that soon.)

filbranden:
CI issue is now resolved with #354.

igrigorik (approved):
👍🏻

PR #358, opened 2022-08-29
Update Docker client image to one with more recent Vim and set up GitHub actions to keep it up-to-date

### LINKED ISSUE #352: Should show <C-End> and <C-Right> instead of their opposite
In my recent solution to the 9v0062e3e96a000000000226 challenge played with the Docker container (the current one, `hettomei/vimgolf@sha256:be79c58788d789129fcf4d87390dd4c61df9d9200d947acde4a7a00698731228`), I used `<C-End>` and `<C-Right>` in insert mode. However, the `Here are your keystrokes:` message before pushing it to the web shows `<C-Home>` and `<C-Left>` instead (and, since I've pushed it, the web site is also showing the solution with these misleading replacement, that's actually how I noticed this). I'm not sure what happened, but the shown sequence of keystrokes is misleading this way, showing the opposite of what I pressed while solving the challenge (I commented this on the solution).

filbranden:
Include a Dockerfile and launcher script in the main repository.

Set up GitHub action to build the image and push it to the GitHub Container Registry, do so weekly in order to keep it updated, with recent updates to the Debian Linux distribution, the Vim binary shipped by Debian, and any changes pushed to the vimgolf gem as well.

The new image is much smaller than the one originally built by Hettomei (40MiB vs 366MiB compressed, relevant for network transfers, and 140MiB vs 1GiB uncompressed, relevant when stored on disk).

The new image is also built for both 64-bit x86 and 64-bit ARM, which makes it also work on the Macbook M1.

I updated the instructions to point to the containers being built in my fork of the repository, where the GitHub Action has been running for a while. Once this PR is merged, the Action should start running here too and container images will be built here as well, so we can update the instructions to point to this one instead.

Fixes #352.

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -1,0 +1,41 @@
+name: Vimgolf Docker client
+on:
+  schedule:
+    - cron: '30 5 * * 2'
+  workflow_dispatch: {}
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out git tree
+        uses: actions/checkout@v2
+      - name: Generate Docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ghcr.io/${{ github.repository }}
+          tags: |
+            type=schedule,pattern={{date 'YYYYMMDD'}}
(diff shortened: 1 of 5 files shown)


filbranden:
Retested this after the latest automated container build this Tuesday and it's working fine. I'll go ahead and merge this!

PR #366, opened 2022-11-15
Replace <C-C><C-C> with <C-C>

### LINKED ISSUE #224: <C-C> gets counted twice when used to exit insert mode
If I enter any challenge, and perform "i<C-C>ZZ", the game counts it as "i<C-C><C-C>ZZ"

dstein64:
Vim intentionally replaces `<C-C>` with `<C-C><C-C>`, which is reflected in the output when using `-W` (https://github.com/vim/vim/issues/11541).

In VimGolf, when a user enters `<C-C>`, the input will count as two keystrokes towards the score, showing `<C-C>` twice in the displayed keystrokes.

This PR updates the code to replace `<C-C><C-C>` with `<C-C>`. This will allow the CLI to not double count `<C-C>` in the reported keystrokes and score. My motivation for handling this in `keylog.rb`, rather than e.g., `cli.rb`, is that the server code would be able to utilize this update as well. However, without updating the database, which [appears to save scores](https://github.com/igrigorik/vimgolf/blob/77c370dc2cf840025847f0974459db6588628298/db/schema.rb#L42-L52), the scores would presumably still be incorrect for _already-submitted_ solutions.

This fixes #224.

### DIFF
--- a/lib/vimgolf/lib/vimgolf/keylog.rb
+++ b/lib/vimgolf/lib/vimgolf/keylog.rb
@@ -9,6 +9,9 @@ module VimGolf
       # Force encoding of solution text. Must match string literals.
       # .force_encoding CHANGES THE ORIGINAL STRING!
       @input = input.force_encoding(Encoding::ASCII_8BIT)
+      # Vim intentionally replaces <C-C> with <C-C><C-C> (Vim #11541).
+      # Convert <C-C><C-C> back to <C-C> (VimGolf #224).
+      @input = @input.gsub("\x03\x03", "\x03")
       @time = time
     end
 
--- a/lib/vimgolf/lib/vimgolf/keylog.rb
+++ b/lib/vimgolf/lib/vimgolf/keylog.rb
@@ -9,7 +9,7 @@ module VimGolf
       # Force encoding of solution text. Must match string literals.
       # .force_encoding CHANGES THE ORIGINAL STRING!
       # Vim intentionally replaces "<C-C>" (\x03) with "<C-C><C-C>" (Vim #11541).
-      # Convert "<C-C><C-C>"" back to "<C-C>"" (VimGolf #224).
+      # Convert "<C-C><C-C>" back to "<C-C>" (VimGolf #224).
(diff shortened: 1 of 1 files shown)


dstein64:
> "My motivation for handling this in keylog.rb, rather than e.g., cli.rb, is that the server code would be able to utilize this update as well."

To elaborate, my thought was that an alternative way to handle this would be to have the CLI fix the issue right when reading the keylog file generated by Vim. Then all new entries from the update CLI wouldn't have the issue.

But I thought that there may be users that don't update their CLI version frequently, and the current version of the PR would properly handle their entry upon submission (although if they were using an older CLI, the wrong keystrokes and score would still be shown when they're using the application).

Upon further thought, I'm less confident in the approach I implemented versus the alternative (having this handled immediately by the CLI when loading the key log script). If you're interested in the type of update in this PR, please let me know if you have any preference one way or the other.

lib/vimgolf/lib/vimgolf/keylog.rb
```diff
@@ -8,7 +8,9 @@ class Keylog
     def initialize(input, time=Time.now.utc)
       # Force encoding of solution text. Must match string literals.
       # .force_encoding CHANGES THE ORIGINAL STRING!
-      @input = input.force_encoding(Encoding::ASCII_8BIT)
+      # Vim intentionally replaces "<C-C>" (\x03) with "<C-C><C-C>" (Vim #11541).
+      # Convert "<C-C><C-C>" back to "<C-C>" (VimGolf #224).
+      @input = input.force_encoding(Encoding::ASCII_8BIT).gsub("\x03\x03", "\x03")
```
dstein64, line 13: I [originally](https://github.com/igrigorik/vimgolf/pull/366/commits/4ae04ee7b1661f4910e30988e349317a221e2c96) had the comment and update on separate lines, after the `input.force_encoding` line, but I moved them both here after a rubocop error from too many lines.

dstein64:
I'm going to close this for now, to explore the alternative approach I mentioned above.

PR #398, opened 2025-12-19
Ignore \xfd\x6d Focus keys

Str8Razor:
Vimgolf already filter out  some vim window focus keys. 
I am adding one more that is missing. 
Without it the output is: "Here are your keystrokes: fd-6d fd-6d:q<CR>". 
This will fix the client side only.  And should be apply to server side as well.

### DIFF
--- a/lib/vimgolf/lib/vimgolf/keylog.rb
+++ b/lib/vimgolf/lib/vimgolf/keylog.rb
@@ -266,6 +266,7 @@ module VimGolf
       "\xfd\x61" => nil, # Focus Gained (GVIM) (>7.4.1433)
       "\xfd\x62" => nil, # Focus Gained (GVIM)
       "\xfd\x63" => nil, # Focus Lost (GVIM)
+      "\xfd\x6d" => nil, # Focus (GVIM)
     })
   end
 end
--- a/lib/vimgolf/lib/vimgolf/keylog.rb
+++ b/lib/vimgolf/lib/vimgolf/keylog.rb
@@ -266,7 +266,8 @@ module VimGolf
       "\xfd\x61" => nil, # Focus Gained (GVIM) (>7.4.1433)
       "\xfd\x62" => nil, # Focus Gained (GVIM)
       "\xfd\x63" => nil, # Focus Lost (GVIM)
-      "\xfd\x6d" => nil, # Focus (GVIM)
+
+      "\xfd\x6d" => nil, # Ignore <OSC>
     })
   end
 end


GamesByFelix:
Is there a workaround for this problem while this is not merged yet?

Str8Razor:
@GamesByFelix 
This patch will just clean keystrokes string that vimgolf shows you after a challenge has been completed.
Until it will be fixed on server side, keystrokes string that is submitted to server should be filtered as well. 
To do this you need to add these 2 lines to /usr/lib/ruby/gems/3.4.0/gems/vimgolf-0.5.0/lib/vimgolf/cli.rb

```
    def upload?(challenge)
+      VimGolf.ui.warn "Cleaning up input keys..."
+      %x( sed -i -e 's/\x80\xfd\x6d//g' #{challenge.log_path} )

      VimGolf.ui.warn "Uploading to VimGolf..."
```

dstein64:
I believe the issue arose from the `<OSC>` pseudo key that was recently added to Vim in [c531501](https://github.com/vim/vim/commit/c53150174859c34e96a65c7435743614890cd335).

Before vimgolf is updated, using a version of Vim without that commit should resolve the issue.

> "This patch will just clean keystrokes string that vimgolf shows you after a challenge has been completed."

~@Str8Razor, I believe your PR fully addresses the issue client-side (I interpreted your comment as suggesting otherwise). That is, I think the patch in your PR can be used to work around the issue without server-side changes (for new submissions; I'm not sure about submissions already made that included the extra keys).~ Update: I see in your PR description that you say "This will fix the client side only. And should be apply to server side as well."

I applied the same type of fix to [dstein64/vimgolf](https://github.com/dstein64/vimgolf) (a vimgolf Python client I implemented) after receiving [Issue #8](https://github.com/dstein64/vimgolf/issues/8).

lib/vimgolf/lib/vimgolf/keylog.rb
```diff
@@ -266,6 +266,7 @@ def each
       "\xfd\x61" => nil, # Focus Gained (GVIM) (>7.4.1433)
       "\xfd\x62" => nil, # Focus Gained (GVIM)
       "\xfd\x63" => nil, # Focus Lost (GVIM)
+      "\xfd\x6d" => nil, # Focus (GVIM)
```
dstein64:
```suggestion

      "\xfd\x6d" => nil, # Ignore <OSC>
```

My understanding is that `\xfd\x6d` is not from GVim focus, but rather from the `<OSC>` pseudo key added in [c531501](https://github.com/vim/vim/commit/c53150174859c34e96a65c7435743614890cd335).

lutalli:
> @GamesByFelix This patch will just clean keystrokes string that vimgolf shows you after a challenge has been completed. Until it will be fixed on server side, keystrokes string that is submitted to server should be filtered as well. To do this you need to add these 2 lines to /usr/lib/ruby/gems/3.4.0/gems/vimgolf-0.5.0/lib/vimgolf/cli.rb
> 
> ```
>     def upload?(challenge)
> +      VimGolf.ui.warn "Cleaning up input keys..."
> +      %x( sed -i -e 's/\x80\xfd\x6d//g' #{challenge.log_path} )
> 
>       VimGolf.ui.warn "Uploading to VimGolf..."
> ```

I can confirm this quick patch works. Thanks!

I use `rbenv` and the file I had to edit was

```
~/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/vimgolf-0.5.0/lib/vimgolf/cli.rb
```

